PR #2, opened 2015-12-04
Copy data format specifications from M. Mayers pdf

joleroi:
* [ ] Event list
* [ ] Effective area
* [ ] Energy dispersion
* [ ] PSF

### DIFF
--- a/source/events/index.rst
+++ b/source/events/index.rst
@@ -3,19 +3,135 @@
 Event lists
 ===========
 
+Event lists are stored in FITS files with three extensions (HDUs).
+
 EVENTS extension
 ----------------
 
-TODO: describe, especially be precise about the non-trivial things that are relevant for analysis:
-
-- TIME
-- Field of view coordinates (columns are optional)
-- Live time, observation time, dead time
-
+The first extension contains characteristic information about each
+event. These information are stored in a FITS binary table. The columns
+are listed in the :ref:`tab_1` table. In addition, a list of header keywords to be contained in each FITS event list is documented in the  :ref:`tab_2` table. Many of the keywords are not necessarily required for an analysis. The information is, however, included as meta data in the event lists to enable instrument-dependent studies and selections of particular observations.
 
 GTI extension
 -------------
 
-`GTI description in OGIP standard <http://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/rates/ogip_93_003/ogip_93_003.html#tth_sEc6.3>`__.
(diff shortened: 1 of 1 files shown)


joleroi:
@cdeil 
@mimayer

Before I copy eveything, we should agree on a format I guess. Do you agree with the current one? (RST tables in a separate section)

mimayer:
what format you mean exactly? For the document or the IRFs?

joleroi:
Sorry, I meant the document.

mimayer:
Ah ok :)
Doesn't really matter to me as long it is readable, searchable and quickly accessible. Of course LaTeX is always nicer but I don't know if it is worth the effort.

cdeil:
I think I would find it a little better if the tables listing the EVENTS HDU header entries and columns would be directly in the section describing the EVENTS HDU.

But that's not super important and very quick to change.

Is the rendering of the output HTML table OK?
(you can run `make html` locally)
In case you're doing a lot of manual LaTeX to RST re-formatting, I'd suggest you give http://pandoc.org/ a try ... I've used it to convert the Gammalib / ctools docs from raw HTML to RST in the past and it was a help (not perfect, but better than manual editing).

@joleroi I have some time to work on this on the weekend.
I guess you wanted to continue on Monday?
(if I don't hear back from you I'll just do other stuff and leave this for Monday)

cdeil:
@mimayer - you now also have the commit bit for this repo.
No need to go through pull requests / review for now.
Let's just add / improve things as we see fit to get something together quickly.

Then we can tag a 0.1 or 1.0 version and after that be more strict with changes, i.e. go through a pull request / review process.

cdeil:
FYI: I've restructured the docs a bit.

There's now an `info` sub-folder with a few pages with background info (that can be references from spec):
http://gamma-astro-data-formats.readthedocs.org/en/latest/

For the IRFs I've started a high-level description of how they are stored in BINTABLE HDUs here:
http://gamma-astro-data-formats.readthedocs.org/en/latest/info/fits-arrays.html#bintable-hdu

I didn't know this before today, but there is some documentation about this way to store n-dim arrays and axis info in FITS BINTABLEs:
http://heasarc.gsfc.nasa.gov/docs/heasarc/caldb/docs/memos/cal_gen_92_003/cal_gen_92_003.html#tth_sEc4

joleroi:
ok no PRs I added the event list description

@mimayer  Is there a reason why the last column in your tables (FITS type) contains somethimes ```TFLOAT (1E)``` and sometimes just ```float```?

mimayer:
@joleroi : I guess not. Probably for header keywords I sticked to just float, while for arrays I used the proper words.

PR #12, opened 2015-12-09
Update index.rst

peger089:
Update to the PSF_TABLE description.

### DIFF
--- a/source/irfs/psf-ogip/index.rst
+++ b/source/irfs/psf-ogip/index.rst
@@ -4,30 +4,32 @@ PSF OGIP
 ========
 
 This is a PSF FITS format we agree on for IACTs.
+This file contains the energy * offset dependent table distribution of the PSF from monte carlo simulations. 
+So far, there is no plan to implement any kind of smoothing before exporting. 
 
 Reference: http://heasarc.gsfc.nasa.gov/docs/heasarc/caldb/docs/memos/cal_gen_92_020/cal_gen_92_020.html
 
-TODO: describe differences / format we use for IACTs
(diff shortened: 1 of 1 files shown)


mimayer:
I have a comment about the name "RAD_LO/HI", which I don't think is very meaningful. At the moment in gammalib, we use "delta" to specify the angle from the source position. We could here use DELTA_LO/HI instead.

Probably mostly irrelevant but just my taste: for the order of the columns I'd prefer: "ENERGY, THETA, DELTA

cdeil:
@mimayer The order of the columns and the names are taken 1:1 from the RPSF OGIP spec.
I've updated the description to explain this more clearly:
http://gamma-astro-data-formats.readthedocs.org/en/latest/irfs/psf_table/index.html

OK to keep these column names?

cdeil:
@peger089 - As discussed, I've changed to column name "RPSF" to match the OGIP spec and fixed the normalisation Sphinx Latex formula:
http://gamma-astro-data-formats.readthedocs.org/en/latest/irfs/psf_table/index.html

This now looks good to me.

@mimayer -- Maybe we should start a new separate issue about consistent axis names (or a table with recommended / in use axis names in the docs)?

mimayer:
Ok, axis names are fine with me. Didn't know they existed already for OGIP. Nevertheless, we shouldn't be guided by OGIP too much since we invent our own stuff :)

cdeil:
I'll make a new issue or pull request about axis names in a few minutes ... checking if we already have inconsistency in some places or not ... agreed we don't have to follow OGIP (but it would be the default I use if there's no other arguments / constraints).

PR #24, opened 2016-01-19
Link GTI to EVENTS via DSS header keys

mimayer:
This pull request proposes to link GTI to a given EVENTS via data sub-space (DSS) header keywords, as used by Fermi-LAT and others.

[edit: title and description made more clear by @cdeil on 2016-02-28]

### DIFF
--- a/source/events/index.rst
+++ b/source/events/index.rst
@@ -23,6 +23,10 @@ keywords to be contained in each FITS event list is also documented. Many of the
 keywords are not necessarily required for an analysis. The information is,
 however, included as meta data in the event lists to enable instrument-dependent
 studies and selections of particular observations.
+Since GTIs are specific to the event list they must be stored in the same file as the event list.
+To ensure consistency, the event list header must contain a reference to the GTI extension. This reference
+is stored using DSS keywords (See `here <https://cta-redmine.irap.omp.eu/attachments/download/1822/SDS07.pdf>`__ (sect 1.13) and `here
+<https://cta-redmine.irap.omp.eu/attachments/download/1823/fits_standard30aa.pdf>`__).
 
 Required Column Names
 ---------------------
@@ -137,6 +141,18 @@ Required Header keywords:
 * ``ALTITUDE`` type: float, unit: km
     * Altitude of array center above sea level (1.835 for HESS)
 
+Required header keywords to link GTI extension
+----------------------------------------------
+
+* ``DSTYP1`` type: string
+    * Type of DS keyword, should be "TIME"
+* ``DSUNI1`` type: string
+    * Unit of GTIs, should be "s"
+* ``DSVAL1`` type: string
+    * GTI type, should be "TABLE"
+* ``DSREF1`` type: string
+    * Name of GTI extension (starting with colon), should e.g. be ":GTI"
+
 Optional header keywords
 ------------------------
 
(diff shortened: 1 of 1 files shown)


cdeil:
IMO using DSS is a major step and needs more motivation / documentation.

Please add a glossary entry for DSS and provide all useful links about what DSS is and how Fermi uses it there:
http://gamma-astro-data-formats.readthedocs.org/en/latest/info/glossary.html
I don't remember where, but there was a post (by Jürgen or you) yesterday with useful links.
Here you only have two links to CTA websites, which aren't visible by many HESS members and the public.

Is DSS something that's supported by any C / C++ / Python library?
Or something one has to re-implement?

Is this the only use case for DSS you have in mind, or are there other good uses for DSS for IACTs DL3?

Isn't there an OGIP convention to link to GTIs?
I think the X-ray instruments do this differently  ... is it a good idea to introduce DSS for IACTs?

Note: I haven't had time to look at how DSS works. My comments might sound negative, but I'm not really opposed to using it. I just would like to know what the gain / scope is before adding such a requirement.

mimayer:
> I don't remember where, but there was a post (by Jürgen or you) yesterday with useful links.
Here you only have two links to CTA websites, which aren't visible by many HESS members and the public.

Links are public on the CTA redmine. Everyone can access them. 

> Is DSS something that's supported by any C / C++ / Python library?
Or something one has to re-implement?

I have no idea what and if there is a python library supporting that. It is used in Fermi and other missions to indicate applied selections to the data.

> Is this the only use case for DSS you have in mind, or are there other good uses for DSS for IACTs DL3?

DSS keywords are used extensively in Fermi and also in ctools to keep records of applied cuts. So this is OGIP convention. E.g. after running ctselect, you have a FITS header containing the following lines 
<pre>
DSTYP1  = 'TIME    '           / Data selection type
DSUNI1  = 's       '           / Data selection unit
DSVAL1  = 'TABLE   '           / Data selection value
DSREF1  = ':GTI    '           / Data selection reference
DSTYP2  = 'ENERGY  '           / Data selection type
DSUNI2  = 'TeV     '           / Data selection unit
DSVAL2  = '0.541551:37.8832'   / Data selection value
DSTYP3  = 'POS(RA,DEC)'        / Data selection type
DSUNI3  = 'deg     '           / Data selection unit
DSVAL3  = 'CIRCLE(83.6333,21.5144,2.5)' / Data selection value
NDSKEYS =                    3 / Number of data selections
</pre>

This is (as far as I know) and OGIP standard and used extensively throughout high level software using FITS data. Currently this is the only way to keep record of applied cuts.

cdeil:
> This is (as far as I know) and OGIP standard

Then please link to the document describing this OGIP standard.

mimayer:
> Then please link to the document describing this OGIP standard.

I am not sure where to find it. Regardless of this, DSS keywords are at least Fermi standard:
https://confluence.slac.stanford.edu/display/ST/Data+SubSpace+keywords

There, it is written that this idea was obtained from Chandra (-> X-ray mission).

cdeil:
@jknodlseder - Do you think the way to link EVENTS to GTI as proposed here (using DSS) is a good idea for the DL3 spec?
Or is DSS something that science tools use, but not DL3 data providers?

cdeil:
This was discussed at the IACT DL3 meeting.

Our proposal:
* We don't add DSS to the spec for now, close this PR.
* We don't support multiple EVENTS and GTI HDUs in one file now, the association is implicit by being in the same file. No referencing procedure needed.
* GTI being in the same file as EVENTS is required in v 1.0 of the spec (I will make a separate PR making this clear)

@mimayer - OK?

mimayer:
sounds good

cdeil:
I'm closing this old issue from 2016 now.

It's still an open question how EVENTS, GTI and IRFs relate.
This will be discussed / decided in CTA in the next ~ year.

Concerning this specific proposal to adopt the Fermi-LAT DSS for CTA, personally I think that might not be a good idea, because it's quite complex and probably not even needed at the DL3 level.
Note that Matthew Wood as former Fermi ST maintainer made a similar comment in the context of the discussion in Gammapy how to do this: https://github.com/gammapy/gammapy/issues/1250#issuecomment-358070916

Issue #15, opened 2015-12-10
Event list spec required columns


cdeil:
The event list spec currently has way too many required columns:
http://gamma-astro-data-formats.readthedocs.org/en/latest/events/index.html

Most science tools analysis only needs energy, ra, dec, time, probably event_ID should also be required.
All the other stuff should be optional.

Also, way too many header keywords are listed as required.

@mimayer - Do you agree? Can you make an update or should I?

mimayer:
@cdeil:
I completely agree. I was drafting the spec from the ctool perspective where things like CORE or XMAX are required to read the events. Of course this needs improvement on the ctools side.
I made an update - what do you think?

I am not sure if the changes can be implemented in ctools quickly. If not, can we agree on exporting optional columns like "CORE" and "XMAX" for the first version?

cdeil:
Better, but I don't see why OBS_ID, MULTIP, DETX, DETY should be required, I think they should be optional. BUNCH_ID should definitely be optional, I don't even know if CTA will have bunches, and certainly it's a detail that shouldn't be present at the DL3 level.

Concerning ctools, well, the first step is to make a change request or drop an email to the ctools mailing list saying that we think something should be changed and in what way and why. @mimayer Can you do that to get the conversation started?

kosack:
+1 for having BUNCH_ID be optional (it's a technical detail of the HESS DAQ implementation, and is not general)

mimayer:
Fine with me: OBS_ID, MULTIP, BUNCH_ID should be optional.
DETX and DETY should be required as these values are important to create background models and other instrument specific stuff, right?

mimayer:
I made the change

cdeil:
> DETX and DETY should be required as these values are important to create background models and other instrument specific stuff, right?

I think they should be optional. They can be computed from RA / DEC.

@mimayer - Let me know if you disagree about DETX and DETY ... if so we'll move the discussion to a separate issue or the ctools or CTA mailing list. Can you run a ctools analysis with an event list that doesn't have DETX / DETY at the moment? I think last time I checked ctobssim didn't even fill it, so if it needs to be present at the moment, it can't be hard to change to make it optional?

mimayer:
> I think they should be optional. They can be computed from RA / DEC.

This depends on the definition of DETX, DETY. Some kind of instrument coordinates are always needed I would say. 

[...]

PR #29, opened 2016-02-26
Update spec of EVENT_ID in DL3 event list

cdeil:
This PR updates the spec of the `EVENT_ID` field for DL3 event lists.

There was already some discussion on this in #15 and (I think) agreement that `BUNCH_ID` should not be in the DL3 event list spec (see  https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/issues/15#issuecomment-163665721).

@mimayer @jknodlseder @cboisson @kosack - Please comment.

### DIFF
--- a/source/events/index.rst
+++ b/source/events/index.rst
@@ -5,6 +5,8 @@
 IACT event lists
 ================
 
+This document describes IACT DL3 event lists.
+
 Event lists are stored in FITS files with two required and one optional
 extensions (HDUs).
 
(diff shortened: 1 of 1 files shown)


kosack:
AGreed, as I said before  `BUNCH_ID` relates to a HESS-specific DAQ implementation, so should not be defined or required. It's better to just say that *instrument-specific columns can be included for bookkeeping purposes, but should not be used in general analysis software*.

cdeil:
Merging this now.
Of course, feedback is still welcome ...

joleroi:
:+1: for a unique event ID, I just wanted to use it for the first time and though there was a bug because it is not unique. I only realize now that you've been discussing this since almost a year ago.

PR #31, opened 2016-04-06
Update index.rst

cdeil:
Let's be more friendly ...


source/events/index.rst
```diff
@@ -72,6 +72,8 @@ at low data levels and is already fixed for existing IACTs, we propose here
 to have an ``EVENT_ID`` that is simpler and works the same for all IACTs at
 the DL3 level.
 
+Hello world!
```
cdeil: I dont' think this is a good change. Rejected.

PR #36, opened 2016-04-21
Generate format spec from YAML file

joleroi:
This is an idea how to represent a format spec in a YAML file. It follows

http://stackoverflow.com/questions/7250659/python-code-to-generate-part-of-sphinx-documentation-is-it-possible

The steps are

* Define spec in YAML file (here:  event_spec.yaml)
* Generate RST include files with generate_includes.py
* Done

The generate_includes step should of course be run as part of the make procedure.

I added an example for the required columns for the event list. The docs look the same as before, the advantage being that the spec is available as YAML file and a tool that takes an event list and check if it fullfils the spec is straightforward to write.

Let me know what you think, if we want to proceed with this I will start by updating the event list spec.


cdeil:
@joleroi - Thank you for this proposal. Obviously, it'll be important to develop IACT DL3 validation tools that data producers and consumers can run to check if some data conforms to the spec or to get a list of errors or warnings.

It's not clear to me how to best set this up though. These validation tools could be implemented close to the spec like you propose here, or as part of larger packages like Gammapy or Gammalib, or in a separate repo.

Specifically with the approach you propose, there's the advantage to have a single reference / no issues with spec and validation getting out of sync, but there's the disadvantage of having a more complex spec writing process. At the moment everyone can click "edit on Github" and fix something or make a change.

For now, I think I would prefer to not continue with this.
(and for Gammapy to continue writing Python code for checking files).
My suggestion would be to keep this PR open for people to comment on, and then to discuss this in one of our monthly IACT DL3 telcons to see what the options are for validation tools and what people want.

@jknodlseder @cboisson @kosack - Thoughts?

joleroi:
@cdeil I didn't necessarily intent to write the validation tools as part of the spec. Actually I am just changing the docs generation process (which I don't think is more complicated now) with the advantage that I have a YAML file defining the spec. You can use that for whatever you want 
* validation
* generation examples file
within or outside the spec

cdeil:
> I didn't necessarily intent to write the validation tools as part of the spec. Actually I am just changing the docs generation process (which I don't think is more complicated now) with the advantage that I have a YAML file defining the spec.

I hear you!
It's a nice idea and we will consider it.

All I'm saying is that I was already happy that people consider to adopt Github, RestructuredText and Sphinx for the spec writing (Note: "consider", we don't have contributions or adoption yet.)
But let's wait for comments from other potential contributors or maintainers of this spec.

cdeil:
This idea to write the spec in machine-readable format or to have a schema was briefly discussed by @woodmd and @cboisson yesterday (see https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/issues/45#issuecomment-245838901 ).

@woodmd likes the idea because he thinks long-term it can be useful.

In the past days I've started to write some schemas in YAML format at
https://github.com/gammapy/gamma-cat/tree/master/input/schemas
and have been using http://json-schema.org/ via https://pypi.python.org/pypi/jsonschema to validate input files.
The idea was to make sure there aren't missing fields or typos in field names or data type errors like putting a string instead of a list of string as value for a given field.

This was more of an experiment. I didn't like json-schema much, I think I'd prefer just writing Python code to validate something.

As for using YAML files just to write the spec in stead of RST (at least not now for schema or validation) : yes, that can be a little useful (e.g. see  [this](https://github.com/gammapy/gamma-cat/blob/master/input/schemas/hgps_cat.yaml) is an example where I did this, and I'm using that to check a FITS file that all columns match and to generate a TeX / PDF with the column description).
But I'm still against it: the YAML->RST is a little extra complexity for spec contributors, and I'm pretty sure that it would quickly evolve into more and more complexity, trying to make those files a schema that can be used for automatic data file validation (e.g. expressing required fields etc.) and we'd either adopt something like json-schema or invent our own little schema language.

Instead I'd prefer to just write a Python package for data file validation in this repo or Gammapy.
Everyone knows how to read and write Python, it's just as simple for simple schemas, and way more powerful for complex things, e.g. data content validation is possible (e.g. no NaN inside safe parameter ranges in IRF arrays).

PR #37, opened 2016-04-24
Binned Likelihood Formats

woodmd:
This is a first draft of the bin-by-bin likelihood format that we discussed back at the time of the PyGamma meeting.  Along with the 2D likelihood format I've also included the format that is currently output by *gttscube* -- a new ballistic tool that we recently made available in an internal ST release (11-00-00).  gttscube extends the concept of the bin-by-bin likelihood to 4 dimensions by performing an SED analysis at every spatial bin in the ROI.

### DIFF
--- a/source/results/binned_likelihoods/index.rst
+++ b/source/results/binned_likelihoods/index.rst
@@ -1,0 +1,113 @@
+.. include:: ../../references.txt
+
+.. _binned_likelihoods:
+
+Binned Likelihood Profiles
+==========================
+
+This page describes formats for bin-by-bin likelihood profiles as
+currently used in some LAT analyses.  
+
+
+2D Binned Likelihood
+--------------------
+
+The 2D binned likelihood is a representation of the profile likelihood
+for the normalization of a source vs. energy.  The 2D likelihood can
+be used in the same way as a traditional SED but has the advantage
+that it retains information about the shape of the likelihood function
+around the likelihood maximum.  The proposed format is a BINTABLE with
+the following columns:
+
+* ``E_MIN`` -- ndim: 1, unit: MeV
+    * Dimension: nebins
+    * Lower edge of energy bin.
+* ``E_MAX`` -- ndim: 1, unit: MeV
+    * Dimension: nebins
+    * Upper edge of energy bin.
+* ``EFLUX`` -- ndim: 1, unit: MeV cm^{-2} s^{-1}
+    * Dimension: nebins
+    * Energy flux of best-fit source normalization integrated over the energy bin.
+* ``DFDE`` -- ndim: 1, unit: MeV^{-1} cm^{-2} s^{-1}
+    * Dimension: nebins
+    * Differential flux of best-fit source normalization evaluated at
+      the bin center.
+* ``NPRED`` -- ndim: 1, unit: counts
+    * Dimension: nebins
+    * Number of predicted counts of best-fit source normalization.
+* ``NORM_ERR`` -- ndim: 1, unit: None
+    * Dimension: nebins
+    * Symmetric error on the source normalization.      
+* ``NORM_ERRP`` -- ndim: 1, unit: None
+    * Dimension: nebins
+    * Positive error on the source normalization.
+* ``NORM_ERRN`` -- ndim: 1, unit: None
+    * Dimension: nebins
+    * Negative error on the source normalization.
+* ``NORM_UL`` -- ndim: 1, unit: None
+    * Dimension: nebins
+    * Upper limit on the source normalization.            
+* ``TS`` -- ndim: 1, unit: counts
+    * Dimension: nebins
+    * Source test statistic as a function of energy bin.
+* ``DNLL_SCAN`` -- ndim: 2, unit: None
+    * Dimension: nebins x nnorms
+    * Delta negative LogLikelihood value vs. normalization.  The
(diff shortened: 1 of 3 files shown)


cdeil:
@woodmd - Thank you!

Here's some first feedback:

- I think it's good to have this as a format in this spec, both for interop between tools, and who knows, maybe some day gamma-ray astronomers will publish spectra this way.
- Can you add one example file for each of the two formats? (ideally something small, a 10k or 100k file, but up to 1MB is OK).
-  For people that see this for the first time, it's hard to grasp the purpose and how it works. Here's some suggestions to make it more accessible / hopefully speed up adoption:
  - Include a image to illustrate it (could be just a PNG you created, or from a Python script that uses the example file and is in the same folder)
  - Change the title? "binned likelihood" is very generic. It seems at least currently, this format only works for spectra, so maybe put "spetrum" or "sed" in the title?
  - Add a description on the exact use cases, pointing out the pros of this format over flux points and also the limitations of using the full likelihood? I guess there are quite a few assumptions, i.e. spectral shape, position and morphology parameters, nearby sources. This could be a new section at the top or at the end of this page
 - I realise this is asking much, and we can merge this as-is also. It's just a request to describe it more to help me and others understand when / how to use it. Actually, I think this could eventually even be a method paper, to propose a better way to publish spectra, but probably it's best to let people use it for a year or two and give feedback and just put it here for now.
- Some details about the format:
  - Don't you need a norm binning, i.e. `NORM_MIN` and `NORM_MAX` like you have an energy binning `EMIN` and `EMAX`?
  - Do you always have all of these things available? I.e. is it worth declaring some columns as optional to  make it easier for tools to produce a valid file, and, if there is redundant info in some fields, to make it clear which info tools can rely on to be present when processing these spectra from multiple instruments?

I can have a closer look at this and try out using this once there are example files available and give better feedback.

woodmd:
Thanks for the comments.  My intention with this PR was to provide a first draft which would be a basis for further discussion.  All of your suggestions for expanding the content of the page (examples, assumptions, use cases, etc.) are good so I can try to work on that over the next few weeks but in the meantime I would propose that we go ahead and do a merge now.

> Don't you need a norm binning, i.e. NORM_MIN and NORM_MAX like you have an energy binning EMIN and EMAX?

I'm not sure what it would mean to define a binning for the normalization parameter.  The spectral model is a pdf in energy (i.e. in the data space) but the normalization parameter is just a scalar.  The `NORMSCAN` array completely specifies the grid of points at which the likelihood was evaluated.  Note that because it's a 2D array the grid spacing can be different in each energy bin.  This is a useful property and allows you to store the likelihood scan much more efficiently than would be possible with a fixed grid.  In my own tests I've found that 5-10 scan points are sufficient to get a relatively accurate representation of the likelihood if you use spline interpolation.

> Do you always have all of these things available? I.e. is it worth declaring some columns as optional to make it easier for tools to produce a valid file, and, if there is redundant info in some fields, to make it clear which info tools can rely on to be present when processing these spectra from multiple instruments?

Yes several of the columns could be declared as optional.  For instance `NPRED` could be dropped without any loss of information about the flux distribution.  On the other hand `NPRED` should always be calculable when performing an analysis on raw data so when possible I think we should try to provide it along with the flux point estimates.

`DFDE` and `EFLUX` are not redundant if you allow for freedom in the spectral distribution within a bin.  Since these should both be calculable from the spectral model alone I would prefer to have them both required.  

Other optional columns could be those for the errors and ULs (`NORM_ERR`,`NORM_ERRP`,`NORM_ERRN`,`NORM_UL95`) since these can all be derived from the likelihood profile.  However having these precomputed is quite convenient and allows the format to be used interchangeably with SEDs (i.e. ignoring the extra information in the likelihood scans).

cdeil:
> The NORMSCAN array completely specifies the grid of points at which the likelihood was evaluated. 

I think I understand now. This wasn't clear to me that `NORMSCAN` contains the grid points at which the likelihoods are evaluated, i.e. I overlooked this:

> Normalization scan values in each energy bin expressed as the ratio with respect to the best-fit normalization.

I can see how that's a good grid in many cases. But what about the cases where the best-fit norm is zero? Doesn't that way to define the grid in normalisation fail in those cases?

That's why I didn't grok it ... I was expecting there to be an equal-log-spaced NORM grid, i.e. a dense matrix.

> I would propose that we go ahead and do a merge now.

OK, go ahead and push the green button!

The description is certainly much better than several other things we have put in, and feedback / iteration is needed anyways for any format, i.e. it has to show up in the docs first.

woodmd:
> I can see how that's a good grid in many cases. But what about the cases where the best-fit norm is zero? Doesn't that way to define the grid in normalisation fail in those cases?

You're right that this leaves some ambiguity when you have a zero best-fit norm.  We've avoided this in the TS cube output by setting the lower bound to a small but non-zero value.  An alternative would be to have another array that defines the reference flux for the normalization scan.

cdeil:
> An alternative would be to have another array that defines the reference flux for the normalization scan.

And yet another possible alternative: make the norm scan grid points absolute, not factors to be multiplied with some best-fit flux.

This seems better to be because then you can support best-fit flux of 0, i.e. the case of negative excess and profiles that correspond to upper limits, and users could generate the likelihood profiles without having to compute the best-fit value, and I find absolute values slightly simpler than relative values.
I.e. keep your clever spacing of norm grid points that samples the likelihoood well, just serialise a different representation.

PR #38, opened 2016-04-25
Create separate page for EVENTS description

jknodlseder:

PR #39, opened 2016-04-25
Change EVENTS and add POINTING

jknodlseder:
Here a new draft for the event list format for tomorrow's meeting


cdeil:
Thank you very much!
I'll review this PR now.

@mimayer @kosack @cboisson - If you have time as well today, please review as well.

As discussed in the IACT DL3 call yesterday, we will merge this tomorrow and further changes / review can then happen in other pull requests.

source/events/events.rst
```diff
@@ -0,0 +1,295 @@
+.. include:: ../references.txt
+
+.. iact_events:
```
cdeil: You changed the label from `iact-event` to `iact_events` and now there's three broken links to this section:
```
/Users/deil/code/gamma-astro-data-formats/source/data_storage/hdu_index/index.rst:67: WARNING: undefined label: iact-events (if the link has no caption the label must precede a section header)
/Users/deil/code/gamma-astro-data-formats/source/info/time.rst:143: WARNING: undefined label: iact-events (if the link has no caption the label must precede a section header)
/Users/deil/code/gamma-astro-data-formats/source/ogip/index.rst:8: WARNING: undefined label: iact-events (if the link has no caption the label must precede a section header)
```
Please fix.
(I don't care what the label is, either version is fine)

source/events/index.rst
```diff
@@ -1,301 +1,16 @@
...
-    * Float part of instrument specific MJD time reference       
-* ``TIMEUNIT`` type: string
-    * Time unit (e.g. 's')
-* ``TIMESYS`` type: string
-    * Time system (e.g. 'TT', 'MJD', 'JD', 'TJD')
-* ``TIMEREF`` type: string
-    * Time reference frame, used for example for barycentric corrections
-      (options: 'LOCAL', 'SOLARSYSTEM', 'HELIOCENTRIC', 'GEOCENTRIC')
+Event lists are stored in FITS files that contain the following
+extensions (HDUs):
```
cdeil: You say that GTI and POINTING are part of the event list here.
For GTI that might be acceptable (but is also controversial), for POINTING it's even less clear to me if that grouping is good, or if POINTING shouldn't be in some section separate from EVENTS (what is always called TECH data in CTA).

Maybe this page would already be a good place to say that POINTING is optional (at the moment)?
Is GTI optional or required at the moment?
I guess you think it should be required and tools should get times from there by default (not the EVENTS FITS header)?
That's fine with me, because it's more flexible and not a big burden to make everyone produce GTI.

source/events/events.rst
```diff
@@ -0,0 +1,295 @@
...
+
+* ``EVENT_ID`` tform: ``1J``
+    * Event identification number at the DL3 level
+      (lower data levels could be different, see note below).
+* ``TIME`` tform: ``1D``, unit: s
+    * Time stamp of event in instrument specific MJD time reference. See the
+      header keywords ``MJDREFI`` and ``MJDREFF`` for the zero point of
+      the reference time.
+      See also the `OGIP event list`_ standard.
+* ``RA`` tform: ``1E``, unit: deg
```
cdeil: To specify the column type (using `RA` as an example for my general point here), I would prefer we use:
```
type: float
```
instead of
```
tform: ``1E``
```

The FITS code ``1E`` is very cryptic ... I have to go look it up every time ... is `E` 32 or 64 bits?

Also -- I don't think we should force 32 or 64 bit here in the spec, it's OK if both are valid.
Personally I prefer 32 bit for RA to save space.

There are a few fields where the number of bits matters, e.g. TIME and there we can write ``int64`` which everyone understands immediately without having to look up FITS number codes.

OK?

source/events/events.rst
```diff
@@ -0,0 +1,295 @@
...
+
+``EVENTS`` extension
+====================
+
+The ``EVENTS`` extension is a binary FITS table that contains an event list.
+Each row of the table provides information that characterises one event.
+The mandatory and optional columns of the table are listed below. In
+addition, a list of header keywords providing metadata is specified.
+Also here there are mandatory and optional keywords.
+The suggested extension name of the binary table is ``EVENTS``.
```
cdeil: suggested -> recommended
(to be consistent with spec terminology elsewhere)

source/events/events.rst
```diff
@@ -0,0 +1,295 @@
...
+    * Event identification number at the DL3 level
+      (lower data levels could be different, see note below).
+* ``TIME`` tform: ``1D``, unit: s
+    * Time stamp of event in instrument specific MJD time reference. See the
+      header keywords ``MJDREFI`` and ``MJDREFF`` for the zero point of
+      the reference time.
+      See also the `OGIP event list`_ standard.
+* ``RA`` tform: ``1E``, unit: deg
+    * Reconstructed event Right Ascension (see :ref:`sky-coordinates-radec`).
+      See also `HFWG Recommendation R3`_ for the OGIP standard.
```
cdeil: We have other places in the spec (e.g. pointing table, observation index table) with RA / DEC columns.
OK to move the `See also `HFWG Recommendation R3`_ for the OGIP standard.` sentence to the general section to avoid duplicating info?
http://gamma-astro-data-formats.readthedocs.org/en/latest/info/coordinates.html

source/events/events.rst
```diff
@@ -0,0 +1,295 @@
...
+* ``DEC`` tform: ``1E``, unit: deg
+    * Reconstructed event Declination (see :ref:`sky-coordinates-radec`).
+      See also `HFWG Recommendation R3`_ for the OGIP standard.
+* ``ENERGY`` tform: ``1E``, unit: TeV
+    * Reconstructed event energy.
+      See also the `OGIP event list`_ standard.
+* ``DETX`` tform: ``1E``, unit: deg
```
cdeil: I'm not happy with this DETX / DETY definition, I don't think anyone outside HESS knows what "nominal system" is, or that it's a good term to propagate. But this was just a side comment, I'm OK to put this in as-is and will (hopefully very soon) make a pull request for FOV coordinates where we can then discuss this.

source/events/events.rst
```diff
@@ -0,0 +1,295 @@
...
+      See also the `OGIP event list`_ standard.
+* ``DETY`` tform: ``1E``, unit: deg
+    * Reconstructed event Y-coordinate in detector system
+      (nominal system, see :ref:`sky-coordinates-fov`).
+      See also the `OGIP event list`_ standard.
+* ``EVENT_TYPE`` tform: ``32X``
+    * Event quality partition.
+
+      .. note::
+         It needs to be discussed whether this should be a mandatory column
```
cdeil: By suggestion would be to make it optional for now, because many IACT event lists have been produced without this. Also, it's not clear to me if 32 bit are sufficient for all use cases.

We could then, after more thought and prototyping, have a proposal that really introduces EVENT_TYPE for IACTs, and at that time also re-discuss whether it should be required in the event list or remain optional.

That said: I don't feel strongly about this. @jknodlseder - If you prefer to leave as-is (required, with this note), OK.

source/events/events.rst
```diff
@@ -0,0 +1,295 @@
...
+         into quality partitions, and the IRFs computed accordingly.
+
+
+Optional columns
+----------------
+
+.. note::
+   None of the following columns is required to be part of an ``EVENTS``
+   extension. Any software using these columns should first check whether the
+   columns exist, and warn in case of their absence.
```
cdeil: This sentence is a bit misleading if taken literally. It says, software should check and warn if the columns aren't there. But science tools shouldn't check and warn if e.g. `COREX` is there.

I think there's agreement that the recommended behaviour of science tools for **all** optional things (columns, header keys, ...) should be:
* Only access what is needed for analysis
* Ignore any extra stuff
* If something needed (or good to have) for a given analysis is missing, give a meaningful error or warning message.

I think this not should be removed here (because it would be the same for any optional section in the spec), and, if you're willing to put it in, add a section in the "general" section of the spec describing this philosophy what optional means (i.e. that it's just stuff we do declare and agree on key names and semantics for the users that do want to use it).

source/events/events.rst
```diff
@@ -0,0 +1,295 @@
...
+    * Reconstructed altitude coordinate of event
+      (horizon system, see :ref:`sky-coordinates-altaz`)
+* ``AZ`` tform: ``1E``, unit: deg
+    * Reconstructed azimuth coordinate of event
+      (horizon system, see :ref:`sky-coordinates-altaz`)
+* ``THETA`` tform: ``1E``, unit: deg
+    * Reconstructed offset from the observation pointing position
+* ``PHI`` tform: ``1E``, unit: deg
+    * Reconstructed position angle from the observation pointing position
+      (position angles are counted counterclockwise from celestial North)
```
cdeil: Again: 👎  on putting info like "position angles are counted counterclockwise from celestial North" here and to move it in the general section. We already have IRFs with THETA as axis and we probably will have IRFs with PHI as axis, i.e. PHI will be used in more than one place, and it's easiest to maintain if the exact definition of PHI is given only in one place and then we reference that place from where it's used.

source/events/events.rst
```diff
@@ -0,0 +1,295 @@
...
+* ``HIL_MSW_ERR`` tform: ``1E``
+    * Hillas mean scaled width error
+* ``HIL_MSL`` tform: ``1E``
+    * Hillas mean scaled length
+* ``HIL_MSL_ERR`` tform: ``1E``
+    * Hillas mean scaled length error
+
+
+Mandatory header keywords
+-------------------------
```
cdeil: Hmm ... I see there's actually significant changes here that will break everything (all events lists produced so far are invalid and no science tools work with the new format specified here).

The most important change is that you've removed RA_PNT and DEC_PNT as event header keywords as key parameters used for analysis and moved them to the more flexible POINTING table.

I see this as the start of introducing required TECH3 that is separate from EVENT3.
I think this is a good move, although the question for me is whether we should do this now, quickly, or take our time and do it later.

Note that if you make this backward-incompatible change, it would probably good as well do also move other info out of the EVENTS FITS header, specifically I think it would be more flexible to have LIVETIME and DEADC info with the GTI info. So I see this as a half-done change to introduce TECH3, with some info left completely or duplicated in the EVENTS header (time info) and others removed (pointing info).

Probably we should either keep RA_PNT and DEC_PNT and backward compat here now, or go more all-in on big changes. @jknodlseder, before we ping others about this, what do you think?

source/events/pointing.rst
```diff
@@ -0,0 +1,56 @@
...
+
+
+``POINTING`` extension
+======================
+
+The ``POINTING`` extension is a binary FITS table that contains for a number
+of time stamps the pointing direction of the telescopes. A *pointing* is here
+defined as the centre of the field of view (or centre of the camera
+coordinates). In reality, all telescopes may point to different positions
+(for example for divergent pointing mode).
```
cdeil: I like your POINTING table suggestion and have no significant comments.
Here in the intro you could mention that the main purpose of POINTING at a given time is to serve as a parameter used in IRF access (i.e. to compute exposure or background).

cdeil:
I'm done with my review. The biggest question for me is about removing the RA_PNT and DEC_PNT header keywords. I'm slightly leaning towards suggesting to leave them in for now (maybe adding a warning that they may be removed in the future in favour of the POINTING table), so that this can go in quickly without upsetting people.

jknodlseder:
Thank you Christoph. I will try to go through your comments as soon as possible (maybe tomorrow morning).

I agree that the POINTING table is something for the future, and for now we can keep the header keywords (at least for the imminent HESS data release). But if we agree in the future that we want to have a POINTING table as proposed, these keywords would not make any sense anymore.

source/events/events.rst
```diff
@@ -0,0 +1,295 @@
...
+* ``MULTIP`` tform: ``1I``
+    * Telescope multiplicity. Number of telescopes that have seen the event
+* ``DIR_ERR`` tform: ``1E``, unit: deg
+    * Direction error of reconstruction 
+* ``ENERGY_ERR`` tform: ``1E``, unit: TeV
+    * Error on reconstructed event energy
+* ``COREX`` tform: ``1E``, unit: m
+    * Core position X of shower
+* ``COREY`` tform: ``1E``, unit: m
+    * Core position Y of shower
```
TarekHC: Maybe add the "reconstructed" also to the core positions? It may be kind of obvious, but more consistent.

cdeil:
@jknodlseder - I and others from HESS would like to work on the IACT DL3 spec this and next week to finish things up for the next HESS FITS production.

Do you have time to address the comments I made above so that this can go in soon and further work on the spec isn't blocked? Or should I continue with this PR starting with what's here?

cdeil:
FYI - We have a first example file for the POINTING table from HESS [here](https://github.com/cdeil/astrometric_checks/blob/master/hess_event_list.fits).

I'd like to work on the spec and HESS data release in the coming days, to have something in master and ready for review for the telcon next Tuesday: https://github.com/open-gamma-ray-astro/2016-04_IACT_DL3_Meeting/blob/master/notes/2016-06-07-IACT_DL3_Telcon.md

@jknodlseder - Do you have time to finish this PR up this week? Or would it be OK with you if I merge this now as-is and continue updating the spec in new PRs?

jknodlseder:
@cdeil Can you remind me about the procedure? Should I just go through all the comments and change the jknodlseder-patch-1 branch or create a new branch?

cdeil:
If you can update this ` jknodlseder-patch-1` branch (by adding new commits and pushing them to Github) that would be ideal.

And please clarify whether this is an old branch / PR or also something that should go in:
https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/pull/38#issuecomment-222605885

source/events/events.rst
jknodlseder: Use `iact-event` now

source/events/index.rst
jknodlseder: Propose to write "Event lists and associated information" (formally you are right that GTI and POINTING are not part of an event list). The text says that POINTING is optional.

cdeil: OK.

source/events/events.rst
jknodlseder: I'd prefer to use the FITS syntax for this, for the simple reason that float in Python is 64 Bits, hence this is very confusion. What counts here is not the Python type, but the TFORM value in the FITS header. That's what you do in a format specification (use the syntax of the format you are working with).

`1E` is 32 Bits in FITS.

source/events/events.rst
jknodlseder: Agree

source/events/events.rst
jknodlseder: Done

source/events/events.rst
jknodlseder: Agree, defer this to FOV discussion

source/events/events.rst
cdeil: My proposal was to write `float` when any float is valid, and `float32` when 32 bit are required and `float64` when 64 bits are required. I'm -1 on writing the cryptic `1E` and putting requirements about 32 or 64 bit on all the fields.

But if you want to keep the FITS codes in this PR, fine.
We can focus on fields and semantics here, and argue the data type spec in a separate PR.
OK?

source/events/events.rst
jknodlseder: Moved to optional for now, but kept 32 Bit (needs more investigation)

source/events/events.rst
jknodlseder: Well, that's not what the text says. It applies to "software **using** these columns", so any software ignoring these columns can stay silent. But I recognised that it does not say what a software that is ignoring these columns should do in case that these columns exist. So I propose:

>    None of the following columns is required to be part of an ``EVENTS``
>    extension. Any software **using** these columns should first check whether the
>    columns exist, and warn in case of their absence. Any software **ignoring**
>    these columns should make sure that their presence does not detoriate the
>    functioning of the software.

Agree that this can be moved upfront, but not sure where you are with the restructuring of the documentation, so I propose to merge this in first and then create an issue to move this note upfront.

source/events/events.rst
jknodlseder: Agree that we can use referencing (but it's important that a reference is inserted for the definition of position angle, which is always counted counterclockwise from celestial North; that's something you do not need to "maintain"). I think we need an issue to collect all information that should be referenced into a single place, but I prefer to do this after merging to keep the mess low.

source/events/events.rst
cdeil: Well, if the software is using the software, it should probably error out if they are absent, not just warn.
I would prefer you remove this note here and then make a separate PR to add it in the General section and we discuss there.
But if you want to leave as-is, I'm also happy to merge and then we'll discuss on a PR where I move it to the General section. :-)

source/events/events.rst
cdeil: @jknodlseder - The idea is to put all general info here:
http://gamma-astro-data-formats.readthedocs.io/en/latest/general/index.html

Assembling it and linking to it from everywhere without duplicating definitions too much is very much work in progress.
I think we'll have to do this a bit incrementally ... it's not realistic to try to get everything right in the first PR.

So I'm happy to merge whatever you put here, then I'll have a look and make further PRs proposing changes in the coming days.

cdeil: PS: if you do want to add links to the general section in this PR you should rebase your branch against latest master ... I did a few updates a few days ago.

source/events/events.rst
jknodlseder: It's indeed a significant change, but formally it's just moving information from a header into a binary table, and we know anyway that we want to go this way. I agree that it requires adaption of some code, but this should be rather minor. As we are working on v0 now, we formally don't have to worry for backward-compatibility, we now define the reference from which we will start. So I think in one or two months from now, it will decrease our headaches, and we will gain tremendously in flexibility (we don't need to use it though for a long time).

I agree that `DEADC` could also be moved to into the pointing table (it should not go into `GTI` as these may be changed by the user). You don't need to do that for `LIVETIME`, you even do not need `LIVETIME` as it can be inferred from `DEADC` and `ONTIME`.

However, these quantities still make sense as average values for the entire event file, fulfilling thus the OGIP standard. I would therefore always keep this information in the header, at least for user information, even if a more fine grained information is used later for TECH3.

source/events/pointing.rst
jknodlseder: It's more than the IRF access, right? It's basically providing the time dependent coordinate transformation for changing between celestial and terrestrial coordinates. Could also be used by tools that do not refer to the IRFs. How about:

> The main purpose of the ``POINTING`` extension is to provide time dependent information on how to transform between celestial and terrestial coordinates.

source/events/events.rst
jknodlseder: Done

jknodlseder:
2 more commits, including all my changes.

source/events/events.rst
```diff
@@ -0,0 +1,290 @@
...
+      If a Good Time Interval (GTI) table is provided, ``ONTIME`` should be
+      calculated as the sum of those intervals. Corrections for instrumental
+      *dead time* effects are **NOT** included.
+* ``LIVETIME`` type: float, unit: s
+    * Total time (in seconds) on source, corrected for the *total* instrumental
+      dead time effect.
+* ``DEADC`` type: float
+    * Dead time correction, defined by ``LIVETIME/ONTIME``.
+      Is comprised in [0,1]. Defined to be 0 if ``ONTIME=0``.
+* ``EVENT_CLASS`` type: int
```
cdeil: EVENT_CLASS should be type: str, no? (the examples you give are strings)

cdeil:
Thanks for making the updates.

There's a ton of required header keys for EVENTS.
Those that aren't needed for science analysis and I think should be made optional.

E.g. OBS_MODE or OBSERVER many aren't available when simulating an event list, but ctobssim should still write valid event lists without having to invent dummy header info, no?

source/events/events.rst
cdeil: > I agree that DEADC could also be moved to into the pointing table (it should not go into GTI as these may be changed by the user). You don't need to do that for LIVETIME, you even do not need LIVETIME as it can be inferred from DEADC and ONTIME.

Ah, this wasn't clear to me.

If you remove RA_PNT and DEC_PNT from the EVENTS header, then POINTING becomes required anyways, because science tools need to know the pointing position to use background models.
I'm OK with this break, but then let's move LIVETIME info to POINTING as well now.
@jknodlseder - Is it OK for you ... e.g. ctobssim will always have to write POINTING tables.

This is getting closer to the Fermi-LAT data model.
Maybe include a reference to this and say that our POINTING is similar to the Fermi-LAT SPACECRAFT?
http://fermi.gsfc.nasa.gov/ssc/data/analysis/documentation/Cicerone/Cicerone_Data/LAT_Data_Columns.html#SpacecraftFile

Also, I think if it's going this way, changing in POINTING from TIME to (START, STOP) is better, because LIVETIME or DEADC is valid for a time interval. Overall moving to START, STOP and saying that the quantities are associated with START isn't a big change.

@jknodlseder - Do you agree that would be a good change to make to POINTING? Assuming yes -- Do you want to do it in this PR or should I do it in a follow-up PR tomorrow?

How does Fermi-LAT handle deadtime information? Is it in the spacecraft file?

To be able to

cdeil:
@jknodlseder - Do you already have an idea how EVENT -- IRF association should work, assuming IRFs are valid for time intervals.
I gather you see neither GTI nor the time intervals in POINTING as the time intervals in which IRFs are valid? Those would be separate and don't have to be considered here when adding POINTING?

source/events/events.rst
jknodlseder: I looked a bit into the Fermi format, and it is interesting, but confusing in the sense that all information is given basically for the START time and only the LIVETIME is given for an interval between START and STOP. In other words, STOP-START serves basically as an ONTIME. I think that implicitly this implies that STOP is never later than the next START.

I believe that it's cleaner (and more readable for a user) to not mix time stamps and intervals in the same data structure. This could in principle by avoided by using DEADC instead of LIVETIME, which would be the deadtime correction to be applied until the next time stamp is reached.

But I'm not sure that we should make here a quick decision. We should try some prototyping to test how this information can be used. Also, I believe that it should be a second order effect, and there should not be drastic variations in DEADC for a typical run duration (but only you know what happens really in data from existing IACTs). So I would propose to keep the header keywords for the moment, and to prototype the addition of DEADC to the POINTING table (there is even nothing bad in having the information in both places, as they apply to different time intervals, hence a software could switch to the more precise short time information if really needed).

However, for the pointing we know that directions constantly change, either in celestial or in terrestrial coordinates. So for sure we need the pointing table.

jknodlseder:
No, I have not thought about this.

But the proposal was not to have intervals for pointings (what would be their meaning for a constantly changing direction?), but time stamps.

jknodlseder:
A string can always stay blank. These are OGIP standard names, so would be good if they are there.

source/events/events.rst
jknodlseder: I guess I took this from Fermi where it's an int (I think it's also a column there), but since we want to have these anyways in a separate file I agree that it should be "str"

source/events/events.rst
kosack: I haven't totally followed this discussion, but I would like to note that putting LIVETIME in the pointing table seems strange to me, as pointing has nothing to do with LIVETIME (which is a quantity associated with EVENTS, not the pointing/tracking system)

cdeil:
There's a few things here that I'm not happy with:
* POINTING table becomes effectively required (to have a `RA_PNT, DEC_PNT` pointing position which is needed as the center for IRF and background evaluation), making all existing DL3 FITS data invalid.
* Use of cryptic FITS data type codes
* A ton of required EVENTS header keywords that aren't needed for analysis (provenance info)

But still I'm merging this now, because it is in other regards a big step forward, and leaving this PR open blocks further work on the spec. I'll try to make new PRs or at least issues on the various points I'd like to change so that we can discuss them in the [IACT DL3 telcon tomorrow](https://github.com/open-gamma-ray-astro/2016-04_IACT_DL3_Meeting/blob/master/notes/2016-06-07-IACT_DL3_Telcon.md).

Thanks, @jknodlseder !

cdeil:
I've fixed some formatting things in master. The new version of the EVENTS and first version of POINTING spec is now visible here:
http://gamma-astro-data-formats.readthedocs.io/en/latest/events/index.html

kosack:
> POINTING table becomes effectively required (to have a RA_PNT, DEC_PNT pointing position which is needed as the center for IRF and background evaluation), making all existing DL3 FITS data invalid

the ``RA_PNT`` and ``DEC_PNT`` columns were always a hack to avoid having to make a separate pointing table or coordinate trasforms anyhow . The pointing table allows one to avoid calculation of ALT/AZ from the ``TIME`` column,  and of course supports any strange non-standard tracking mode.  Recall in the case that the telescopes are tracking normally such that RA/Dec of the center of the FOV is fixed, the pointing table can be optional. FOr that reason a tracking mode should be put in the header, and if mode='tracking', the pointing table can be used if it is found, or ignored and computed on the fly if not.

Issue #35, opened 2016-04-15
Energy true/estimated binning in "aeff_2d"


TarekHC:
Hi all, 

Aren't we currently "imposing" effective area vs true energy to have the same binning as the effective area vs estimated energy? Isn't that completely misleading?

I would propose to add another pair of columns defining the true energy binning (even if its identical). 

I edit it further: The energy column is always labelled as "Energy axis". I would propose to always specify to which energy we are referring to. Specially in the case of IRFs.

jknodlseder:
Good point.

But if we do that, the question is whether we should not simply store that information in another extension (or have the energy type as a keyword)?

Another question if we should formally support this way of approximating the energy dispersion?

TarekHC:
Hi @jknodlseder!

One quick question: Does ctools currently use both effective areas? 

We are providing the energy dispersion, so going from TENERGY -> RENERGY is trivial. Couldn't we just provide the effective area vs true energy?

Again, we could go even further. We could multiply the mig. matrix with the effective area in bins of estimated energy to create the gamma ray acceptance as a function of both true and estimated energy. That would allow us to reduce (marginally) IRF3 volume (as the mig matrix would actually be within the effective area matrix).

Although this last idea (coming from Abelardo here at IFAE) would probably not be similar to current standards... And probably our current approach is more organized.

jknodlseder:
No, ctools does always use the true effective area and handles energy migration correctly.

I'm wondering whether we loose in fact information in the approach you propose, because you need to do a convolution and not a multiplication.

TarekHC:
If we only use effective area vs true energy, then why should we add the other one? I would remove it then. 

I think there is no loss of information. In fact, I think MAGIC in some steps of the analysis does something similar. From the integral of each slice you get the effective area of that bin, and scaling the gaussian with that value will give you back the mig. matrix. But again, that is probably a bit misleading also...

cdeil:
@TarekHC - Thanks for bringing up the issue.

This is a tricky point and we need to find a good solution and document it well.

How does Fermi-LAT support the analysis case where energy resolution isn't applied by the ST in their IRFs?

If AEFF versus ETRUE is used, this results in a shift of ~ 5 - 10 % in flux (or equivalently energy scale) for constant AEFF and power-law spectrum, depending on the energy resolution and source spectral index. So Fermi-LAT must do *something* to handle this, no?

woodmd:
[...]

PR #43, opened 2016-04-28
Modified effective area format. EFFAREA_RECO column moved to a new HDU.

TarekHC:
As discussed in issue #35 , I added some changes to the effective area format. 

Essentially, I removed the  ``EFFAREA_RECO`` column, and described the effective area vs estimated energy as a separate HDU.

The format may be definitely improved. I'm not sure if the "``aeff_2d`` format" should refer to both effective areas, or they are "different formats".

I also modified the python macro that generates the FITS file example, updating it.


cdeil:
@jknodlseder - Do you have time to review this?

From my side, I think the main comment would be that there should be a declarative way to tell which AEFF a given HDU is. I guess this means that the `HDUCLAS2` header keyword should be different for the two types of AEFF?

source/irfs/effective_area/index.rst
```diff
@@ -54,4 +52,34 @@ sources.
...
     * On region radius for point-like observations
 
+Although not a requirement, the recommended ``EXTNAME`` keyword is "EFFECTIVE AREA".
+
+.. _aeff_reco_2d:
+
+Effective Area vs reconstructed energy
+--------------------------------------
+
+Although not required, it is generally useful to store the effective area as a function of the reconstructed energy,
```
cdeil: > Although not required, it is generally useful

Can you rephrase here?

Both (AEFF, EDISP) and the combined AEFF_RECO are in use for current IACTs and both are useful.
Whether CTA could just pick one and never need the other is unclear to me ... my guess would be no, i.e. both will be used.

* Using AEFF, EDISP is more precise, but it requires enough MC stats to have an EDISP with acceptable noise, and it's slower to use in likelihood fits for science tools.
* Using AEFF_RECO is less precise, requires less MC stats and is faster to use in likelihood fits for science tools.

Maybe it's OK to extend the description of this format to explain this briefly why they both exist and when to use which one as DL3 producer (i.e. roughly the short summary I wrote here)?

TarekHC:
Thanks @cdeil ,

>From my side, I think the main comment would be that there should be a declarative way to tell which AEFF a given HDU is. I guess this means that the HDUCLAS2 header keyword should be different for the two types of AEFF?

As the HDUCLAS2 was not present there before, I decided to not include it. I will simply add it to both effective areas description.

> Maybe it's OK to extend the description of this format to explain this briefly why they both exist and when to use which one as DL3 producer (i.e. roughly the short summary I wrote here)?

Ok, verbose mode on. I will commit those changes today.

TarekHC:
Ok, done. I tried to implement all comments. I also re-structured the text a bit. Let me know if you think I should change anything else.

cdeil:
@TarekHC - Thanks!

Assigning @jknodlseder for review / decision.
I'm going on holiday for two weeks and am out of time. 
Feel free to merge, don't wait for comments from me.

TarekHC:
@cdeil @jknodlseder , we are all extremely busy with Kashiwa... So there is really no rush. Just review it whenever you find the time.

cdeil:
@TarekHC - Can you please rebase this pull request against master? (Github shows that it can't be merged because there is a conflict).

@jknodlseder - Do you have time to review this PR this week?

From my side it's OK ... the main point is that it splits `EFF_AREA_RECO` in a separate HDU now and there are `HDUCLAS2` header keywords to declare what kind of response it is.

TarekHC:
Yes, no problem. I'll take care of that. Although I will probably wait for jurgen's feedback first, as new conflicts may arise.

cdeil:
@TarekHC - Can you please rebase this PR now?

Background info: I'd like to merge this (and the other open PRs) now, so that further work on the spec isn't blocked. We'd like to clean up several details for the next HESS FITS production that will then be used for the test data release.

TarekHC:
Oh, sure. I'll try to do it today then. Cheers!

TarekHC:
Ok, I think the conflicts were solved.

PR #44, opened 2016-04-29
Binned Likelihood Formats v2

woodmd:
This is a second draft of the binned likelihood format specifications.  I've added some material to the top of the page with a general discussion of how the binned likelihoods can be used and some of their limitations.  The current implementation keeps the `NORM` columns with fit values, errors, etc. expressed in units of a reference model amplitude defined by the `REF` columns.  The `REF` columns can be used to convert the `NORM` columns into different units by doing an element-wise multiplication.  This approach avoids duplication in the table columns (e.g. having error or upper limit columns for each unit).

There are still few things that are currently missing like a sample file and images for the TSCube format but I plan to add those in a future commit.  I also haven't worked out exactly which columns should be designated as optional so any feedback on that would be welcome.


cdeil:
@woodmd - Thank you very much for this!
If you have the bandwidth my advice would be to consider writing a method paper on this in addition. This might be the future of publishing spectra (individual papers, but even more catalogs) in gamma-ray astronomy, no?
A way that allows to convey more info about the likelihood than global model fit parameters / errors or the flux points, in a way that allows joint likelihood fits with other likelihoods in this format (with strong caveats, of course).

To simplify the review of this pull request, I've put the rendered HTML version corresponding to this pull request [here](https://www.mpi-hd.mpg.de/personalhomes/deil/public/gamma-astro-data-formats/issue-44/results/binned_likelihoods/index.html).

I don't have time to review this at the moment.

@woodmd - Let us know if you'd like us to review and discuss this in detail now, or merge very soon and file issues later in case there are suggestions for enhancement or change.

@zblz @giacomov @registerrier @jjlk - Can you have a look at this, please?

woodmd:
The basic methodology has been described in the recent LAT dSph papers (see e.g. [Ackermann et al. 2014](http://adsabs.harvard.edu/abs/2014PhRvD..89d2001A)).   I'm not sure if a dedicated publication is really necessary but we could think about it.  Unfortunately I'm completely oversubscribed with other projects so I don't think I'll have time to take the lead on this.

With respect to the PR since you don't have time to review this now maybe it would be better to go ahead and merge and move the discussion into an issue thread.

cdeil:
> With respect to the PR since you don't have time to review this now maybe it would be better to go ahead and merge and move the discussion into an issue thread.

OK, but let's keep it open for 1 day.
There might be typos or other little issues that are best fixed directly.
If there's major comments we'll still merge and move those as notes to a new issue as future work.

cdeil:
Merging this now. Thank you!

woodmd:
Thanks.  Further comments should be left in #45.

PR #47, opened 2016-05-12
Add int code for unknown data quality

lmohrmann:
This PR adds an integer code of `-1` for unknown data quality for the observation index files.


cdeil:
Thanks!

I think this is uncontroversial ... merging this now.

If someone has comments on this, please still put them here. If there's some issue with this we can still change it.

PR #51, opened 2016-06-01
Use reference time in obs-index

lmohrmann:
Make the times in the obs-index file consistent with those in the events files.


cdeil:
Thanks.

Note that there's changes to the time columns for events in #39 by @jknodlseder .
Probably we'll want to update the obs index format to again work the same way once that EVENTS change is merged.

But for now, I'm merging this PR as a step in the right direction.

Issue #22, opened 2016-01-18
Add time binning and lightcurve spec


cdeil:
It would be nice if we had a spec for:
- defining time bins as input to analysis tools
- store light-curves.

References:
* http://gamma-astro-data-formats.readthedocs.org/en/latest/results/light_curves/index.html
* https://cta-redmine.irap.omp.eu/boards/6/topics/213?r=229#message-229

Issue #45, opened 2016-05-02
SED and Binned Likelihood Formats


woodmd:
This is an open thread to collect comments and feedback on the proposed formats for [Binned Likelihood Profiles](http://gamma-astro-data-formats.readthedocs.io/en/latest/results/binned_likelihoods/index.html).  Since these formats are effectively just SEDs with extra information in each bin it would probably be a good idea to try to share conventions with the [Flux points format](http://gamma-astro-data-formats.readthedocs.io/en/latest/results/flux_points/index.html).  @zblz Did you have any thoughts on this?

Here are a few open questions:

- Do we want a fixed 95% CL for the UL columns?  I saw that the naima format specifies the confidence level with a header keyword so maybe we should do something similar.

- Do we need a column to specify the bin center?  The current format only specifies the lower and upper edges (`E_MIN` and `E_MAX`).  Differential quantities are assumed to be evaluated at the geometric center of the bin but we could add a bin center column (`E_CTR`) that would allow the evaluation point to be explicitly defined. 

- Which columns do we want to make optional?  The current specification calls for five different unit conversion factors (`DFDE`,`E2DFDE`,`FLUX`,`EFLUX`,`NPRED`).  I mainly did this for convenience so that one doesn't need to know anything about the underlying spectral model to convert between units.  Of these `E2DFDE` could clearly be optional (or maybe dropped entirely) since it can be trivially computed from `DFDE`.  `NPRED` provides additional information that can't be rederived from the other four unit columns but since it may not always be available it should probably also be optional.

- Do we want a column for exposure?  This can be inferred from `FLUX` and `NPRED` but having it as an independent column could be useful in some circumstances.

cdeil:
> Do we need a column to specify the bin center?

Yes, please.

E.g. most HESS spectra have been published with the points not at the log bin center, but using this weighting method:
https://gammapy.readthedocs.io/en/latest/tutorials/flux_point/index.html

I'm not saying that particular method is useful, but I think there are use cases where people want to specify the center and it's easy to support.

In Gammapy we currently  have this class for flux points and just call this column ENERGY:
https://gammapy.readthedocs.io/en/latest/api/gammapy.spectrum.DifferentialFluxPoints.html

For me,  `E_CTR` or whatever you propose is fine ... we'll just adapt Gammapy to whatever is put in this spec.

--- 

I do have some thoughts on the other points you raise, but don't feel strongly either way about any of them, so won't comment for now.

[...]

PR #61, opened 2016-06-10
Add light curve spec (a first draft)

cdeil:
This PR adds a first draft for a lightcurve spec.
(Related: #22)

I've started to write this up because @cnachi and @helen-poon are starting to work on time analysis methods in Gammapy (see https://github.com/gammapy/gammapy/pull/564) and we need to have some I/O format for simple light curves.

My proposal here is that we start with something small, as simple as possible, and merge this very soon.
Then people working on lightcurves can extend and change the format over the coming months as use cases are implemented and a good data model / format emerges.

@zblz @jknodlseder @woodmd, all - can you please have a look today?

EDIT: I added a task list for myself to keep track of action items based on discussion below:
- [ ] Add script to produce examples files (and commit example files)
- [ ] Mention that this can be stored in FITS, but also e.g. ECSV
- [ ] Change time scale from TT to UTC
- [ ] Mention that lower-energy light curves are supported to a limited extent and that e.g. `erg/cm2/s or Jy` units are allowed.
- [ ] Allow flexible energy units as well, e.g. `erg` or `TeV` or `MeV` is OK.
- [x] Add assumed spectral index header key? -> No. Wait for SED spec to see how they do it.

### DIFF
--- a/source/results/index.rst
+++ b/source/results/index.rst
@@ -13,3 +13,4 @@ other codes (e.g. to check results, combine results in one plot, ...).
 
    flux_points/index
    binned_likelihoods/index
+   light_curve/index
--- a/source/results/light_curve/index.rst
+++ b/source/results/light_curve/index.rst
@@ -1,0 +1,81 @@
+.. include:: ../../references.txt
+
+.. _lightcurve:
+
+Light curve
+===========
+
+.. warning::
+
+   This is a first draft proposal.
+   The format is likely to change.
+   Please give feedback!
+
+Introduction
+------------
+
+The light curve format specified here is a simple table where each row
+represents the ``FLUX`` measurement of a given astronomical object at a given
+``TIME``.
+
+* If you have multiple light curves, e.g. from different instruments or energy bands,
+  you can use multiple tables in this format.
+* To specify the ``TIME``, ``MJD`` values in the ``TT`` system should be used.
+* The flux is allowed to be a differential or integral flux, as long as the units
+  are given correctly. All of these are valid:
+    * ``cm^-2 s^-1`` (integral flux)
+    * ``m^-2 s^-1 TeV^-1`` (differential flux)
+    * ``cm^-2 s^-1 MeV^-1`` (differential flux)
+* For integral fluxes, the energy band should be specified via the ``E_MIN`` and
+  ``E_MAX`` header keywords. For differential fluxes, the ``E_REF`` header keyword
+  should be used.
+* For upper limits, the ``FLUX`` field should be set to ``NaN``.
+
+
+Required columns
+----------------
+
+* ``TIME`` type: float64, unit: s
+    * MJD time in TT system
+* ``FLUX`` type: float, unit: flexible
+    * Flux (see notes on units above)
+
+Optional columns
+----------------
+
+* ``TIME_MIN``, ``TIME_MAX`` type: float64, unit: s
+    * Start and end of time bin (MJD, TT)
+* ``FLUX_ERR`` type: float, unit: flexible
+    * Flux error (1 sigma) (see notes on units above)
+* ``FLUX_UL`` type: float, unit: flexible
+    * Flux upper limit (at ``UL_CONF`` level) (see notes on units above)
+
+
+Required header keywords
+------------------------
+
+None.
+
+Optional header keywords
+------------------------
+
+* ``E_MIN``, ``E_MAX`` -- type: float, unit: TeV
+    * Energy band of the flux measurement
+* ``UL_CONF``
+    * Confidence level of the upper limit given in the ``FLUX_UL`` column.
+* ``OBJECT`` type: string
+    * Observed object
+* ``TELESCOP`` type: string
+    * Telescope
+
+Example files
+-------------
+
+TODO: Make an example FITS file in this format.
+
+Existing light curve providers (not in this format though):
+
+* http://vobs.magic.pic.es/fits/
+* http://fermi.gsfc.nasa.gov/ssc/data/access/lat/2yr_catalog/ap_lcs.php (uses MJD only)
+* http://fermi.gsfc.nasa.gov/ssc/data/access/lat/msl_lc/ (uses MET only)
+* http://heasarc.gsfc.nasa.gov/W3Browse/fermi/fermilasp.html (uses MET and JD)


cdeil:
@cboisson @jlenain - I forgot to CC you on this.

- Does this cover common use cases where gamma-ray spectra are shared?
- Is there a preference among AGN people how exactly times are stored (which format: JD or MJD, and which scale: TT or UTC or ...)?

jlenain:
Hi @cdeil 

It already reads well. There is however one important aspect which I see missing: one would want to know what was the underlying spectral assumption under which the light curve was computed, if any. E.g., did it assume a power-law spectrum, or a log-parabola ? This will change the flux values, at least when talking about integral fluxes, and also differential fluxes given at another energy than the decorrelation energy.

Concerning the time format, I don't think there is specific preference, but usually MJD in UTC is quite common.

cdeil:
> It already reads well. There is however one important aspect which I see missing: one would want to know what was the underlying spectral assumption under which the light curve was computed, if any. E.g., did it assume a power-law spectrum, or a log-parabola ? This will change the flux values, at least when talking about integral fluxes, and also differential fluxes given at another energy than the decorrelation energy.

I agree this would be nice to add to the spec.
But I'd like to not attempt this in this first pull request.

The question of how to store that information is already under discussion here for the SED format:
https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/issues/45#issuecomment-218313905

The only short-term, small addition I could think of would be to add an "assumed spectral index" header keyword, to cover the very common power-law case. Should I do that or is it too limiting to be useful?

> Concerning the time format, I don't think there is specific preference, but usually MJD in UTC is quite common.

At the moment I have TT, because that will be the recommendation for other formats (EVENTS and OBS_INDEX). But I don't think consistency with those is important ... so I'll change to UTC here if it's more common.

jlenain:
> The only short-term, small addition I could think of would be to add an "assumed spectral index" header keyword, to cover the very common power-law case. Should I do that or is it too limiting to be useful?

That may be too limiting, indeed. Would a string header keyword as 'SPEC_TYPE' be appropriate ?

> At the moment I have TT, because that will be the recommendation for other formats (EVENTS and OBS_INDEX). But I don't think consistency with those is important ... so I'll change to UTC here if it's more common.

No problem to stick to TT, as long as we know which scale is used :-)

cdeil:
> That may be too limiting, indeed. Would a string header keyword as 'SPEC_TYPE' be appropriate ?

Yes, maybe. But let's wait for the discussion in #45 to conclude and get applied to SED, and then just re-use it for lightcurves here.

> No problem to stick to TT, as long as we know which scale is used :-)

While looking around at existing light curves this morning I found this:
http://fermi.gsfc.nasa.gov/ssc/data/access/lat/2yr_catalog/ap_lcs.php
They don't say if their MJDs are TT or UTC. The difference is about a minute:
```python
>>> from astropy.time import Time
>>> t1 = Time(54697, format='mjd', scale='tt')
>>> t2 = Time(54697, format='mjd', scale='utc')
>>> (t1 - t2).sec
-65.18399999999538
```
So it doesn't matter much, but still, @woodmd or someone else from Fermi-LAT, maybe it's worth clarifying that point there?

zblz:
Do we want to cover lower-energy lightcurves? Allowed unit should then include, e.g., erg/cm2/s or Jy.

cdeil:
> Do we want to cover lower-energy lightcurves? 

I'd say yes, it's a goal to try to support that as well.
But the primary goal are gamma-ray light curves.
So I would try to make the format such that it just works for lower-energy light curves where possible, but not extend it with other domain-specific info or common things (like wavelength or frequency columns instead of energy, or optical bandpass information in the FITS header, ...).

> Allowed unit should then include, e.g., erg/cm2/s or Jy.

Yes, they should.
I'll add some more info concerning allowed units in the next update.

woodmd:
I'm not sure about allowing the ``FLUX`` column to represent quantities with different dimensionality.  When parsing the file how does one determine what type of quantity ``FLUX`` represents?  I would suggest using separate column names for different representations.  For instance following the convention we use in the SED spec you could allow any of the following column names for the normalization:

``FLUX``: Integral flux.
``EFLUX``: Energy Flux.
``DFDE``: Differential Flux evaluated at ``E_REF``.

Columns for errors/ULs would then be prepended with the given normalization column name (``EFLUX_ERR``, etc.).  

You might consider making ``E_REF`` a column rather than a header keyword.  For instance if you want to evaluate the differential flux at the decorrelation energy then ``E_REF`` would need to change from bin to bin in the light curve.

I would suggest adding optional columns for asymmetric errors (``FLUX_ERRP``, ``FLUX_ERRN``) and scans of the profile likelihood (``DLOGLIKE_SCAN`` and ``FLUX_SCAN``).  Having the likelihood information can be useful if you want to merge low significance time bins.

Regarding TT vs UTC I believe in the STs we use the TIMESYS header keyword to designate which scale is used.  However I don't know if this keyword is part of the FITS standard or something specific to the STs.

cdeil:
My preference would be to have a single `FLUX` column which can represent different quantities:
* differential flux
* integral flux
* energy flux

Tools can look at the unit to figure out which quantity it is, or we could declare this via a `fluxtype` header keyword. And often tools don't have to figure this out automatically, I think usually you have an astronomer looking at the lightcurve data and then deciding how to use it.

I find this format a little simpler to specify and to process in code than the alternative that @woodmd proposed: separate columns `FLUX`, `EFLUX`, `DFDE` where at least one out of the three has to be present.

If you have `EFLUX` and `DFDE`, you'll also need to add the corresponding error columns, because sometimes someone just wants to publish a lightcurve as energy flux and energy flux error and that should be a valid lightcurve. I also think that putting `FLUX`, `EFLUX` and `DFDE` gives a false sense that this is complete ... there's many use cases where people want to store e.g. count, rate or exposure columns also. If we give recommended names for optional columns like `EFLUX` or `DFDE`, maybe we should also include `COUNTS`, `RATE` and `EXPOSURE`?

I'm not 100% what I and we want here.
@woodmd , @zblz , @jlenain and others that might be interested in this, if you could comment again that would be helpful.

I'll then try to implement a new version that hopefully satisfies everyone's use cases at least from this small group participating in this discussion. Then the community can start using it and propose extensions in the coming months / years.

woodmd:
A light curve is basically just a sequence of SEDs with a single energy bin or evaluation point so I think it would make sense to try to use the same conventions here as in the current SED format specs.  To that end I'd be happy to make adjustments to the SED specs to be more consistent with what you use in this spec.

In the SED spec we currently have a ``NORM`` column to represent the amplitude, ``NORM_XXX`` columns to represent errors and ULs on the amplitude, and ``REF`` columns that can be used to convert the ``NORM`` columns into different quantities (flux, differential flux, counts,etc. ).  The motivation for this scheme was to allow the SED to be easily converted between different representations without incurring a lot of duplication in the columns -- e.g. the NORM_ERR column can be used to calculate the error on flux, differential flux, energy flux, counts, or any other quantity for which there is a corresponding ``REF`` column.

Here would be my proposal for a common convention that could be used in both the lightcurve and SED formats:

* Use ``NORM``, ``NORM_ERR``, ``NORM_ERRP``, ``NORM_ERRN``, ``NORM_UL`` to represent  the amplitude, errors, and ULs.  Asymmetric errors and ULs would be optional.
* All ``NORM`` columns have a UCD header keyword that identifies what type of quantity is represented (flux, energy flux, differential flux, rate, counts etc.).  
* If the UCD header is empty or undefined then the user can use the ``REF`` columns to convert the ``NORM`` columns into the respective quantity (e.g. as in the present SED formats).
* If the UCD header is defined then the ``REF`` columns can be optionally provided as a means of converting the ``NORM`` columns into a different representation.  For instance if the ``NORM`` column has a flux UCD and there are ``REF`` columns with flux and energy flux UCDs, one can multiply ``NORM`` by ``REF_EFLUX``/``REF_FLUX`` to convert the normalization to energy flux.

In this scheme the ``REF`` columns would be an optional part of the spec that can be provided if you want multiple representations to be stored in a single table.  What I'm not completely sure of is what convention we should use for writing UCD keywords to a FITS file.  In the astropy discussion thread that Karl linked it appears that both  ``TBUCD`` and ``TUCD`` are used in different applications.

eacharles:
Hi All, Matt just pointed me at this discussion.  I'm a bit concerned that the conventions your are proposing for the persistent data will result quite a bit of extra complications for people trying to use the files.  

First off, I don't think you really want to overload the name FLUX to refer to several different kinds of things with different units, etc..., and then have people to refer to a different bit of data in to figure out what it is.   There are just too many ways that can cause confusion.    I think it would be cleaner and simpler to name the column by what it is (FLUX, EFLUX, DFDE or E2DFDE).   

If you want to do something clever in code to allow a class to be initialized using either a flux or and energy flux, or a differential flux and then to provide you with the correct conversions, that is fine and almost certainly very useful, but I it think that it is better to avoid that sort of ambiguity in the persistent representation.   (In part b/c you have no idea about what code people are going to use to open the persistent files in question).  

A second point is that the business with the NORM was originally intended for a specific case, where we are deriving curves of likelihood w.r.t. the normalization of a reference spectrum.   Since we are working bin by bin, and we are using the same spectra over and over again (i.e., for 10s of thousands for pixels), it was convenient to save several different representations of the reference spectrum, and then express the likelihoods w.r.t. to those.   In that particular case, since we are fitting the likelihoods bin by bin, working with the normalization and storing the reference spectra avoids the need to carry around additional information about the exact spectral form.   And, so long you use energy bins that are relatively narrow, it means that you can pretty much trivially convert between the photon flux, energy flux or differential flux.  

On the other hand, I can imagine that there are lots of cases where what you want is simply a spectrum, or perhaps a set of spectra, maybe with different binning, etc...   In those cases you don't really gain anything by defining a reference spectrum.    For those cases I think it would be easier and clearer just to write EFLUX or FLUX or DFDE or whatever to the persistent file.   

In terms of defining a standard for what goes in the persistent file, I would focus on defining what different things mean, if they are present, and leave it up various developers / users to decide what needs to go into a specific file, or what is needed to create an object of a particular class.   

I.e., it would be useful to have a list like:

EMIN, EMAX -> Bin Edges (Used for integral fluxes)
EREF   -> (Used for differenttial fluxes)
 
FLUX    -> Integral Photon Flux
EFLUX   -> Integral Energy Flux
DFDE     -> Differential Photon Flux 
E2DFDE  -> Differential Photon Flux scaled by E^2
XXX_ERR -> Uncertainties on any of the above

REF_XXX     -> Reference value for a particular type of flux
NORM         -> Normalization w.r.t. to the reference fluxes
NORM_ERR -> Uncertainty on normalization w.r.t. the reference fluxes

Then people can decide if what they want is
EMIN,EMAX,EFLUX,EFLUX_ERR  or
EREF,DFDE,DFDE_ERR  or
EREF,REF_DFDE,NORM,NORM_ERR   or 
EMIN_EMAX,REF_EFLUX,NORM,NORM_ERR

Depending on the particular use case.  I suspect that trying to enforce that any specific data must be present is just not going to work all that well.    Of course you probably want code that can convert back and forth between REF_EFLUX*NORM and EFLUX, and you might want some convention that EREF is sqrt(EMIN,EMAX) unless specified otherwise.    And I suspect that we want to write a binned spectrum class that is smart enough to figure out what is present in a file and initialize itself accordingly, but I think you want to keep the number of required items in the persistent files to the absolute minimum.

Regards,

-e

jknodlseder:
I would also prefer a more explicit format where the information content can be inferred from the column name. Hence +1 for Eric's and Matthew's suggestions.

I think the time interval over which the flux was measured should also be indicated (you cannot measure a flux at a point of time, only over an interval of time). This can be done by adding a column for the time bin width (e.g. ``TIME_WIDTH``), or by specifying ``TIME_MIN`` and ``TIME_MAX`` columns. I think I'd prefer the later.

For the time reference, I would just add the usual OGIP keywords to define the time reference (``MJDREFI``, ``MJDREFF``, ``TIMEUNIT``, ``TIMESYS`` and ``TIMEREF``). This is pretty standard, and provides the required flexibility.

cdeil:
For now, just a few comments / replies:
- OK, I agree to drop the idea sometimes using the `flux` column for different kinds of fluxes depending on a `flux_type` header keyword. 
- Instead, let's do what others suggested and use different columns for different quantities as in the [SED](http://gamma-astro-data-formats.readthedocs.io/en/latest/results/flux_points/index.html#sed) spec.
- Previous art: I recently became aware of http://adsabs.harvard.edu/abs/2010A%26A...524A..48T which defines a "simple lightcurve format" and a collection of data in that format here:  https://astro.desy.de/gamma_astronomy/magic/projects/light_curve_archive/index_eng.html
I don't think we should adapt their field names, internal consistency with our `SED` format is more important. But I do think our format should be able to represent those published lightcurves and like their format be simple in the sense that it's possible to just store `time` and `flux`, i.e. make almost all fields optional. This is the same point I'm advocating for SED here: https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/issues/45#issuecomment-244329080
- I'm not working with lightcurves much myself. If someone else wants to continue this branch or just start a new one, and propose a new lightcurve spec taking what was discussed here into account, I'd be very happy. If this doesn't happen, I'll finish this up in a few weeks.

cdeil:
@jknodlseder made this suggestion:

> I think the time interval over which the flux was measured should also be indicated (you cannot measure a flux at a point of time, only over an interval of time). This can be done by adding a column for the time bin width (e.g. TIME_WIDTH), or by specifying TIME_MIN and TIME_MAX columns. I think I'd prefer the later.

> For the time reference, I would just add the usual OGIP keywords to define the time reference (MJDREFI, MJDREFF, TIMEUNIT, TIMESYS and TIMEREF). This is pretty standard, and provides the required flexibility.

I think this is a good suggestion!

If I understand correctly, the common case of just having "MJD values" can be represented as a subset of that format via these header keys:
```
MJDREFI =  0
MJDREFF =  0
TIMEUNIT= 'd'
```

@jknodlseder - Is this correct?

---

Assuming we go this way: many existing lightcurves don't explicitly contain these header keywords, 
and probably tools will assume defaults when reading those instead of flat-out refusing to read them. @jknodlseder @woodmd  Is it possible to agree on a recommended set of values in Gammalib / Fermipy / Gammapy? I guess it's problematic to assume defaults because for lightcurves "days" are common, but for events "seconds", and both "UTC" and "TT" are common.

We're re-writing time-related code in Gammapy at the moment, and in gamma-cat we'll have to reformat the lightcurves anyways to go convert to any format we agree on here now. So at the moment I don't have a strong opinion yet on what we put, I just want to have a written-down format to refer to.

---

For the flux-related columns: I think there's agreement to just re-use what @woodmd has defined for SEDs. Apologies for initially suggesting something different and causing the long discussion here.

cdeil:
Thank you all for your feedback and the discussion here!

I did implement a minimal page for an LC spec in #113, following the suggestions made here. As far as I can tell, there were no major disagreements left, it was pretty clear what to put. I hope #113 was OK.

I'm closing this old PR now.

PR #65, opened 2016-09-16
Updates to SED Fomat

woodmd:
This PR is a comprehensive rewrite of the SED format.  See #45 for previous discussion.  The intention is to create a flexible format that can handle many possible use cases:
* Representing published integral or differential flux points.
* Supporting many different normalization representations (flux, energy flux, differential flux, counts, etc.).
* Supporting inclusion of likelihood vs. flux profiles (e.g. as in the Likelihood SED format).

A few open issues/questions:
* Currently the format lacks a specification for writing UCDs to the column metadata.  I think earlier we discussed using `TUCD`.
* I copied the current UCDs from what is used in the 3FGL.  However it seems like the existing VO UCD types don't really map very well to the quantities that we typically use in gamma-ray astronomy.  For instance there doesn't seem to be an existing keyword that maps to differential photon flux so instead we use `phot.flux.density`.  The more logical choice would be something like `phot.counts.density`.  There also doesn't seem to be anything in the existing VO spec which could be used to represent photon counts.
* Currently all column names and header keywords are upper case but in some ways I find lowercase more natural.  Is there a strong motivation to use upper case?


cdeil:
@woodmd - Thanks!

I probably will only have time to review this in detail after the HESS collaboration meeting next week.

@zblz, @cboisson, anyone - If you have time to have review this, please do!

woodmd:
@cdeil Any idea when I could expect some feedback on this?  In particular it would be useful to know if people are happy with the existing columns names and definitions.

We're putting together a new public release of the Science Tools and I would like to incorporate the format changes from this PR.

cdeil:
@woodmd  - Would a deadline for comments a week from now (Oct 19th) be OK for you or do you want this decided and merged sooner?

Everyone interested in spectra - Please review and comment this PR!

woodmd:
I think we should go ahead and merge now.  It's easy enough to submit additional changes with another PR and merging it now makes it easier for people to look over the current proposal.

cdeil:
@woodmd - I'll review this today.

Please split the UCD issue mentioned above out into a separate issue. I don't think we'll solve this quickly, and comments in closed issues will be forgotten.

+1 to use lowercase field names as much as possible (preferably everywhere).
Feel free to put lowercase here, and then I'll do a separate issue changing other specs where we still use uppercase and having discussion on that in a separate issue.

cdeil:
One more thought: is it clear / should we specify how SED fitting tools could / should compute the likelihood exactly when flux points are given? I.e. whether a `chi^2` fit should be done using a Gaussian PDF in `y=flux` or `y=log(flux)`?
Is it even clear for currently published flux points like in the Fermi catalogs?

woodmd:
Sure I can open another issue about the UCDs.

I think the default presumption should be that the likelihood is normal in flux.  I can't really think of a circumstance where a log-normal likelihood would arise.  In the low statistics limit the likelihood is normal near the peak and exponential at higher fluxes but it's not clear that there is a simple way of parameterizing that behavior (which was part of the motivation for having the explicit likelihood values).

woodmd:
I implemented the switch to lowercase.  Is there anything else that needs to be addressed before merging?

cdeil:
Looks good to me.
(I only reviewed it briefly.)
I'm merging it now.

Thank you very much!

Everyone -- if you write or read SEDs, please try to adopt this format and let us know if it works for you or if something should be changed or improved.

cdeil:
For reference -- the new specs are here:
* http://gamma-astro-data-formats.readthedocs.io/en/latest/results/flux_points/index.html
* http://gamma-astro-data-formats.readthedocs.io/en/latest/results/binned_likelihoods/index.html

cdeil:
The header keys are still uppercase in the spec:
http://gamma-astro-data-formats.readthedocs.io/en/latest/results/flux_points/index.html#header-keywords
Can we please go all lower-case?

gbelanger:
Yes, it seems to me that they should either all be upper case, or all lower case, and it looks like most have been defined as lower case, with just a few exceptions like RATE, SED_TYPE, NORM_REP, etc. For consistency, as well as ease of reading, I also think it would be better to have them all lower case.

cdeil:
@woodmd - Do you agree? Assuming yes, should I make a commit or will you?

woodmd:
I'm fine with making the header keywords lowercase as long as FITS allows it.  I don't recall seeing a lower-case keyword before but I don't know if this is just a convention or an actual restriction of the FITS format.  I'd also want to double-check first that the Fermi STs FITS interface library permits writing lower-case header keywords.

Note that we did decide to make one exception to the lower-case column name convention when the energy bounds and reference model are written in a separate EBOUNDS HDU (as in the tscube format).   Since the counts cubes written by the Fermi STs write an EBOUNDS HDU with upper case column names we decided it would be better to keep the same convention for SEDs for better compatibility with tools that can already parse the EBOUNDS HDU.  I'll try to commit these changes tomorrow (as well as fix a few other typos I found).

kosack:
CFITSIO will automatically change them to upper-case, from what I understand (not sure about astropy.io.fits) since the FITS standards doesn't guarantee the case is preserved.  That may not apply to column names, however, just to header keywords.

The FITS standard does state that column names must not rely on case (they must be case-insensitive), so you cant have a column named `E_0` and `e_0` simultaneously, and you are not guaranteed that the dictionary you get back of column names is in the same case as when you wrote it.

see the discussion here for example. https://community.lsst.org/t/fits-and-lowercase-header-keys/1184/3

woodmd:
I've opened #70 with a few additional fixes/tweaks to the format.

I confirmed that astropy.io.fits automatically casts header keywords to upper case so for now I guess we are stuck with upper case.  With respect to the column names I wasn't aware that FITS standard doesn't preserve case.  I guess this must be a difference between the specification vs. implementation since astropy.io.fits does appear preserve column name case when reading/writing a table to FITS.

kosack:
I think the column names are actually case-insensitive but case-preserving like filenames in macOS.

cboisson:
From the astropy doc: "Although keyword names are always in upper case inside the FITS file, specifying a keyword name with Astropy is case-insensitive, for the user’s convenience. "

PR #70, opened 2016-11-06
SED spec updates

woodmd:
This PR includes some improvements and fixes for the SED spec:
- Fix sample files to reflect column name changes.
- Add sample files in HDF5 format.
- Add new sample files for SED likelihood.
-  Typo fixes.


woodmd:
I don't think any of these changes are controversial so I'm merging this now.

cdeil:
@woodmd - Thanks!

FYI: we're starting to use the SED format in gamma-cat, see [example](https://github.com/gammapy/gamma-cat/blob/master/docs/data/sources/000024/gammacat_000024_2015Sci...347..406H_sed.ecsv).
I still have to check if we encode upper limits correctly.

Issue #75, opened 2016-11-23
Clarify SED upper limits


cdeil:
@woodmd - It's not clear to me yet how to store upper limits in the SED format.

The spec (see [here](http://gamma-astro-data-formats.readthedocs.io/en/latest/results/flux_points/index.html#error-columns)) doesn't explicitly say what values to fill for upper limits and the example files don't contain upper limits.

In Gammapy we currently don't use an `UL` column, but instead a `NaN` value in the down error column to store an UL (see [here](https://gammapy.readthedocs.io/en/latest/api/gammapy.spectrum.DifferentialFluxPoints.html)).
I think (but didn't double-check, the Fermi-LAT catalogs also don't have an `UL` column, but declare ULs by `NULL` value in the lower error. This is just to mention examples, I'm not saying this has an influence on what we specify here.
@zblz proposed an extra bool `ul` flag column to say which point is which, instead of "magic values" (see [here](http://naima.readthedocs.io/en/latest/dataformat.html)).

At this point I don't have an opinion on which UL point serialisation option is best. Is there an advantage (e.g. simpler storage or processing) to not introducing an `UL` colum, but to re-use the lower / upper error columns?
If we keep the UL column, should we require some special values like `NaN` or `NULL` in the other "flux" columns, or should their content be undefined / irrelevant for ULs?

Are there use cases for "lower limits", i.e. should we treat "upper limits" in a special way or just have a system that works for symmetric confidence intervals?

@adonath or @joleroi or @registerrier @giacomov  - Do you have a preference here?

Maybe this?
```
flux flux_err flux_ul
val val nan # not an UL
nan nan val # is an UL
```
Or this, to still allow storing the maximum likelihood estimate, even for UL?
```
flux flux_err flux_ul
val val nan # not an UL
val val val # is an UL
```

@woodmd - Do you have time to make a pull request for this soon?
Or should I make it?
Please say your preference in any case.

(We need this for `gamma-cat`, so if possible, I'd like to decide and implement this in the spec and update the examples to show UL cases in the next days.)

woodmd:
In fermipy we compute ULs, values, and TS for all bins and then allow the user to choose the between the UL and value according to the TS.  In my view having both UL and value for every bin is useful since there are circumstances where you may want to use the UL even if the bin is significant or use the value for low significance bins.  This also gives extra flexibility to choose a different significance/TS cut for selecting between the value and UL representations.

[...]

PR #76, opened 2016-11-26
Clarify conventions for empty values and ULs

woodmd:
This PR addresses #75 by with the following changes/updates:

- Clarify how null/empty values should be dealt with.
- Add a `ul` column to flag whether a measurement should be treated as an upper limit.
- Update sample files to include upper limits.


source/results/flux_points/index.rst
```diff
@@ -80,6 +78,33 @@ error and upper limit columns are:
 * ``flux_ul`` : Upper limit with confidence level given by ``UL_CONF``
   header keyword.
 
+A row may have a value and any combination of upper limits and errors:
+
+.. code::
```
cdeil: I think it should be
```
.. code-block:: none
```
instead of
```
.. code::
```
(see http://www.sphinx-doc.org/en/latest/markup/code.html)

(used three times, I'll just leave one comment)

source/results/flux_points/index.rst
```diff
@@ -80,6 +78,33 @@ error and upper limit columns are:
...
+A ``nan`` value should be used for empty or missing data such as a bin
+for which there is an upper limit but no value and vice versa, e.g.
+
+.. code::
+
+   e_ref dnde dnde_err dnde_ul
+   1000.0 1.0 0.1 nan
+   3000.0 nan nan 1.16
+   
+The ``ul`` column is an optional boolean flag that can be used to
```
cdeil: Maybe call the column `is_ul` meaning "is an upper limit"?
That's more explicit than `ul`, which people might mistake as the ul value column if they don't read the spec carefully.

@woodmd - Just a suggestion, OK to leave as-is if you prefer.

source/results/flux_points/index.rst
```diff
@@ -120,8 +145,9 @@ SED format defines an *SED Type* string which is set with the
 columns that must be present in the SED.  The SED types and their
 required columns are given in the following list:
 
-* ``diff_flux_points``: ``e_ref``, ``dnde``, ``dnde_err``
-* ``int_flux_points``: ``e_min``, ``e_max``, ``flux``, and ``flux_err``
+* ``dnde``: ``e_ref``, ``dnde``
+* ``flux``: ``e_min``, ``e_max``, ``flux``
+* ``eflux``: ``e_min``, ``e_max``, ``eflux``
```
cdeil: +1 -- I like this "minimal" definition of SED types, not mentioning error or UL fields.

source/results/flux_points/index.rst
woodmd: Good catch.  Fixed.

source/results/flux_points/index.rst
woodmd: Yea I think I prefer ``is_ul`` as well.

woodmd:
> So there's two ways to write SED files for the published, archival SED data when ULs are present, right?
I can just use nan and not put the ul bool column, like you do in diff_flux_points.ecsv, and science tools should understand this.
Or I can in addition add the ul column, to be more explicit, and then science tools can use that instead of inferring the ULs from the nan values, right?

Yes you can use either ``nan`` values or the bool column to designate which rows are ULs.  The bool column is really only necessary if you want to explicitly flag a measurement as an UL when both the UL and two-sided interval are defined.  When parsing an SED one should look for the bool column first and otherwise fall back to checking for ``nan`` values.

cdeil:
> Yes you can use either nan values or the bool column to designate which rows are ULs. The bool column is really only necessary if you want to explicitly flag a measurement as an UL when both the UL and two-sided interval are defined. When parsing an SED one should look for the bool column first and otherwise fall back to checking for nan values.

Thanks for clarifying / confirming. Makes sense.

I'd even be +1 to add these extra sentences of description to the spec.

woodmd:
Ok done.  Merging now.

cdeil:
@woodmd - Thank you very much!

cdeil:
I'm re-formatting the SEDs in gamma-cat now to follow this UL convention.

One more thing I noticed: the SED example files don't have a `UL_CONF` set in the header. @woodmd - Can you please add it?
And can you please clarify in the spec for the `UL_CONF` header key if the value put should be in the range `0 ... 1` or `0 ... 100` (i.e. in unit `%`)?

cdeil:
FYI: I've started to use `UL_CONF` in the range `0 .. 1` for the SEDs in gamma-cat.

@woodmd - Is this OK with you?
Should I make a PR to make this clear in the spec and add it to the example files?

woodmd:
Sorry for not getting back to you about this.  Yes UL in the range 0 to 1 is the convention I had intended.  Please go ahead with a PR.

cdeil:
I clarified that the UL_CONF range is 0 to 1 in 19cd1dfce5d6367489d774d4cf1dae839ec9f57c directly in master.

It's already present in the example files. I must have been looking at an old version of the files.

PR #77, opened 2016-12-10
Add E^2 dnde to SED format?

cdeil:
While implementing support for the SED format in Gammapy (see https://github.com/gammapy/gammapy/pull/810), @adonath noticed today something.

The spec currently defines `eflux` as an integral quantity, per energy bin:
> eflux : Energy flux (integral of E times dnde) between e_min and e_max. Dimensionality: energy / ( time * area )

I missed that and had assumed that `eflux` is defined as the value commonly shown in "SED" plots, the differential quantity `(e_ref ^ 2) * dnde`.

Concretely in gamma-cat, there's this published HESS SED that was given as `(e_ref ^ 2) * dnde` in the paper, and I put the values incorrectly as `eflux = (e_ref ^ 2) * dnde`:
https://github.com/gammapy/gamma-cat/blob/1815c4f14d3e369356f0e55d7885cd4792b6b051/input/papers/2016/2016arXiv160908671H/tev-000096.ecsv

Here's one other published HESS SED from a year ago where the values given are `(e_ref ^ 2) * dnde` and the column is called `e_Flux` there:
https://www.mpi-hd.mpg.de/hfm/HESS/pages/publications/auxiliary/auxinfo_HESSJ1018

So we think that there's a need to be able to store `(e_ref ^ 2) * dnde` and propose to add a new column `e2dnde` for that. That would satisfy our use cases and would only be a small change.

@woodmd - OK?
Or would you prefer some other solution?
Can you change the spec or should I make a PR?

### DIFF
--- a/source/results/flux_points/index.rst
+++ b/source/results/flux_points/index.rst
@@ -50,10 +50,13 @@ and ``e_max`` columns.  Differential representations are quantities
 evaluated at a discrete energies defined by the ``e_ref`` column.  The
 supported :ref:`norm_columns` are:
 
-* ``dnde`` : Differential photon flux at ``e_ref``. Dimensionality:
-  photons / (time * area * energy)
-* ``rate`` : Photon rate between ``e_min`` and ``e_max``. Dimensionality:
-  photons / time.
+* ``dnde`` : Differential photon flux at ``e_ref``.
+  Dimensionality: photons / (time * area * energy)
+* ``e2dnde`` : Defined as ``(e_ref ^ 2) * dnde``.
+  A commonly published and plotted differential flux quantity.
+  Dimensionality: energy / (time * area)
+* ``rate`` : Photon rate between ``e_min`` and ``e_max``.
+  Dimensionality: photons / time.
 * ``flux`` : Photon flux (integral of ``dnde``) between ``e_min`` and
   ``e_max``. Dimensionality: photons / ( time * area )
 * ``eflux`` : Energy flux (integral of E times ``dnde``) between
@@ -153,6 +156,7 @@ columns that must be present in the SED.  The SED types and their
 required columns are given in the following list:
 
 * ``dnde``: ``e_ref``, ``dnde``
+* ``e2dnde``: ``e_ref``, ``e2dnde``
 * ``flux``: ``e_min``, ``e_max``, ``flux``
 * ``eflux``: ``e_min``, ``e_max``, ``eflux``
 * ``likelihood``: See :ref:`likelihood_sed`.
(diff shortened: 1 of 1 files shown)


cdeil:
@woodmd - I've attached a commit here adding `e2dnde`. Please review.

> For the spec maybe we could at least recommend that people use dnde over e2dnde when dealing with differential fluxes.

I was trying to type a sentence with that recommendation, but then felt like there should be a "because ..." at the end with a motivation, and came up empty.

In TeV publications, about half of the SEDs are `dnde`, half are `e2dnde`.
I prefer `e2dnde`, because the TeV spectra then aren't so steep and better visible, and because it's what's usually (always) used when comparing with radio / x-ray / GeV data.
So if I could choose a recommendation, it would probably be to always publish `e2dnde`, not `dnde` values.

So for now I didn't add a recommendation.

@woodmd - If you want to add one, can you please put it?

Maybe "... because SED models are usually given as `dnde(e)` formulae" or "... because `dnde` is closer to the naive way to measure a differential flux as `dn / de / dt / dA`" would be valid motivations?

My main point here was that it should be possible to store `e2dnde` values in SED files, and that's possible now.

woodmd:
My view is that scaling by E^2 is simply a visual aid (analogous to using a different stretch on a colormap) and doesn't actually reflect any real difference in the underlying measurement.  For instance in CR physics it's quite common to see plots where spectra are scaled by other powers of energy (2.7 or 3.0) in order to better see deviations from some global spectral shape but the measurement itself is often given in tabular form as an unscaled differential flux.  

If you have a mix of `e2dnde` and `dnde` then every code that does SED fitting will need to check first which of the two is present and then convert to some consistent internal representation.  I guess this extra overhead isn't really such a big deal given how complex the format already is.  Maybe once we have more experience trying to make SED files interoperable we'll come to a different opinion on this.  In any event feel free to go ahead and merge this PR.

cdeil:
There is this one nice feature of `e2dnde` that it makes different parts of the SED comparable in how much energy is emitted. E.g. if `e2dnde` in X-rays is 42x times as high as `e2dnde` in gamma-rays at some energies `e_x_ray` and `e_gamma_ray`, then the X-ray emission contains 42x times more energy for a given width in log energy. I think this feature might be why `d^2 dn/de` or `v F_v` is the most popular for broad-band SED plots.

For now, I've added one sentence to the spec just to clarify that the two are equivalent and why we added the redundant `e2dnd2`:

> The ``dnde`` and ``e2dnde`` representations are equivalent. We define both here, because both are in common use for publications and plots.

Also, FYI: in `gamma-cat` I've chosen to convert the cases of `e2dnde` is used in the paper to `dnde` in the output folder, i.e. currently all TeV SEDs in that archive are given in `dnde` to users.

PR #78, opened 2017-03-10
HEALPix SkyMap and SkyCube Formats

woodmd:
This PR is a first draft documenting the HEALPix data formats currently used in the Fermi STs.  Also included are proposals for new formats that allow one to define energy-dependent HEALPix map order.  I plan to update the page at a later point to include sample files.


cdeil:
@woodmd - Thanks!

Some feedback:
* I wonder if it's a good idea to put "ST HEALPIX" in the section title. I mean, wouldn't the idea / hope be that other codes use the same conventions and also that this somehow can grow to have something that potentially the Fermi ST don't use? So even if it's just a dump of what Fermi-LAT ST have now, I think maybe drop ST from the section title, and just keep the explanation of the origins and current use in the introduction?
* The intro could link to the FITS standard document and explain how the formats defined here relate to the HEALPIX spec there. Is it correct that for things that are in the standard (like ORDERING or other keys) the FITS standard is followed here, and this is a domain-specific extension? Or are there differences wrt. the FITS standard due to historical or other reasons? Maybe we could ask  whoever wrote the HEALPIX part of the FITS standard to have a look at this page and send feedback?
* In the intro you mention `HPX_CONV` and other formats. To be honest, I don't understand how these other formats relate to the formats you define here, or what I should do as a tool implementer to support these other formats. Can you / is it worth to make that clearer?
* What if I want to store multiple maps? So for Fermi ST you always make separate FITS files? The alternative would be to have extra "value columns" in the HDUs, but that is not done, right? Also having two maps in the same FITS file isn't easily possible because the EBOUNDS / ENERGY HDU reference is just by name, not by linking from the HEALPIX header, right? I think that's fine as-is, but worth pointing out in the intro to people that if they have multiple maps, they must use multiple FITS files in this scheme.
* [EBOUNDS](http://gamma-astro-data-formats.readthedocs.io/en/latest/skymaps/healpix/index.html#ebounds-table) table is the same or very similar as for maps / cubes with WCS? Maybe move the description of those HDUs before or after the WCS and HEALPIX maps pages and reference it from both?
* All the example files and descriptions of the four formats use `EBOUNDS`, only the intro mentions `ENERGIES` for differential quantities. I presume all formats work for integral and differential? Is there a way to make this clearer? (again, maybe by splitting out EBOUNDS and ENERGIES description on it's own page and reference it frequently?)
* What are the limitations here (e.g. on ORDERING) if the common 64-bit ints are used for indexing? Maybe mentioning at least the rough angular resolution one can deal with here in the intro would be useful?
* For energy bin index `BIN_IDX` is used consistently. Personally I would find `ENERGY_IDX` or `IDX_ENERGY` better, to avoid confusion with the pixel and global index, and because for differential maps it's not really an energy "bin", but a "plane" or "channel". I'm not sure if this is even up for debate: is the term here used in FITS files (column names, header keys) already and fixed, or is it only used in the spec and can be chosen freely?

cdeil:
One key question is how COORDSYS should be filled and interpreted.

You put this:
```
COORDSYS, type: string
Map coordinate system. Can be CEL (celestial coordinates) or GAL (galactic coordinates).
```

It seems that others use the codes "C", "G":
https://healpy.readthedocs.io/en/latest/generated/healpy.fitsfunc.write_map.html
http://healpix.jpl.nasa.gov/html/idlnode58.htm
https://github.com/astrofrog/reproject/blob/691aea031fd995b095cd66d22316bff1112389c4/reproject/healpix/utils.py#L9

Is this the situation?
* Fermi ST require exactly COORDSYS = "CEL" or "GAL" and write that.
* Other tools will write COORDSYS = "C" or "G" and might or might not work with "CEL" or "GAL" (depending on whether they look at the first character or whole string)?
* If the header keyword is missing, is "CEL" always assumed to be the default by everyone or will Fermi ST fail to read such a file?

There is also some ambiguity what CEL and GAL are exactly:
* Probably CEL is FK5 J2000 in the Fermi ST and as ICRS by some others (e.g. Astropy reproject)
* There might be differences about exactly which Galactic frame is used (see [here](http://gamma-astro-data-formats.readthedocs.io/en/latest/general/coordinates.html#galactic) what we defined so far in the spec).
The differences are small (~ 10 arcsec), but as a library or tool implementer you still need to know what to put.

Probably since there are a decade or more of files and tools around already, we cannot just define whatever we like and agree on one convention that all tools understand (or can we)?
So my suggestion here would be to move this key point concerning `COORDSYS` to a separate section with a bit more explanations how readers / writers should behave, instead of just listing `COORDSYS` as a header keyword and leaving it up to tool implementers or users to figure this out.

woodmd:
> I wonder if it's a good idea to put "ST HEALPIX" in the section title. I mean, wouldn't the idea / hope be that other codes use the same conventions and also that this somehow can grow to have something that potentially the Fermi ST don't use? So even if it's just a dump of what Fermi-LAT ST have now, I think maybe drop ST from the section title, and just keep the explanation of the origins and current use in the introduction?

Yes I think the idea is that these specifications would eventually become an official standard and would not be tied to the STs (although we would try to support them in the STs as much as possible). However until we agree on what the format should contain I thought it was better to refer to these as the ST HEALPix format.  

> The intro could link to the FITS standard document...

I haven't carefully reviewed the official FITS HEALPix spec so I don't know which parts are already in the spec.  At least some of the header keywords (``ORDER``, ``NSIDE``, etc.) seem to be universally used but I've never seen a document where they are officially defined.  I found this [document](https://fits.gsfc.nasa.gov/standard40/fits_standard40draft1.pdf) from a google search but it doesn't seem to have any reference to HEALPix header keywords.

> In the intro you mention HPX_CONV...

Yes I need to improve and expand this section.  The idea was to document other conventions for column names so that packages can support these older formats (probably only for reading).

> What if I want to store multiple maps? So for Fermi ST you always make separate FITS files?

You can store multiple maps as long as they have the same geometry in the energy dimension.  An example would be the ST source map file which contains maps for all sources in the ROI model but only a single EBOUNDS HDU.  I agree that it's not really ideal to split the information across HDUs but unfortunately that's the convention that was chosen so changing it at this point would be pretty painful.  We could modify the format to support inclusion of maps with different energy binning in the same file by adding a header keyword that defines the energy hdu name.

> EBOUNDS table is the same or very similar as for maps / cubes with WCS? Maybe move the description of those HDUs before or after the WCS and HEALPIX maps pages and reference it from both?

I agree that it would be better to document this in a single place.

> All the example files and descriptions of the four formats use EBOUNDS, only the intro mentions ENERGIES for differential quantities. I presume all formats work for integral and differential? Is there a way to make this clearer? (again, maybe by splitting out EBOUNDS and ENERGIES description on it's own page and reference it frequently?)

Yes I agree that the explanation could be improved here.

> What are the limitations here (e.g. on ORDERING) if the common 64-bit ints are used for indexing? Maybe mentioning at least the rough angular resolution one can deal with here in the intro would be useful?

I haven't given that much thought.  I doubt that we will ever need to go beyond maps of order 12 (~0.01 deg pixels / 100M pixels) so for 64-bit integers I don't think we need to worry.  We probably should require all indices to be 64-bit just so that we don't run into problems when dealing with higher order maps.

> For energy bin index BIN_IDX is used consistently. Personally I would find ENERGY_IDX or IDX_ENERGY better, to avoid confusion with the pixel and global index, and because for differential maps it's not really an energy "bin", but a "plane" or "channel". I'm not sure if this is even up for debate: is the term here used in FITS files (column names, header keys) already and fixed, or is it only used in the spec and can be chosen freely?

We are already using these column name conventions in the STs.  However we are open to making changes if there is a good argument for doing do.  Changing the ST implementation at this stage shouldn't be too disruptive since the HEALPix functionality is not yet widely used and we always include logic to make the STs backwards-compatible with the old convention.

> One key question is how COORDSYS should be filled and interpreted.

Yes I agree it makes more sense to have a dedicated page on COORDSYS and just link to it here.  I believe the Fermi STs use FK5.  Going foward perhaps we could adopt the following convention:

* `C`, `CEL`, `FK5` : Celestial (FK5)
* `ICRS` : Celestial (ICRS)
* `G` or `GAL` : Galactic

cdeil:
CTA has a requirement that point source localisation must be 10 arcsec at low energy and 5 arcsec at high energy. So Ideally additional errors from e.g. binning effects in a HEALPIX analysis would be ~ 1 arcsec. Maybe there could be a statement / rough estimate somewhere at the top, what resolution is available with 64 bit HEALPIX indices (maybe assuming 10 or 100 energy bands)?
I'm not sure how to exactly calculate it.

For TIME we had a similar question, i.e. whether a single 64 bit float is good enough to support all CTA use cases, and concluded that for DL3 and high-level analysis the answer is yes. To document this I wrote a section here stating what time precision we roughly get with the EVENTS format we have here: http://gamma-astro-data-formats.readthedocs.io/en/latest/general/time.html#precision

I have the same question about precision when it comes to COORDSYS. If I remember correctly, the maximum difference between ICRS and FK5 J2000 for any point in the sky is 15 arcsec. So it doesn't matter for almost all gamma-ray analyses, but for very few (e.g. the Galactic center and Crab localisation), it does. Plus, it's good to have a common, written down understanding of what coordinate system is used, to avoid having small pointless differences and discussions between different codes.

Unfortunately, as far as I can see, this isn't defined in the FITS standard.

In the HIPS paper I found this:
http://www.aanda.org/articles/aa/full_html/2015/06/aa26075-15/aa26075-15.html


> For astronomical purposes the generic HEALPix spherical coordinate system (θ,φ as in Górski et al. 2005) is mapped to astronomical coordinates so that the positions of the pixels have precisely defined positions on the sky (Fig. 1). For HiPS we consider only two astronomical coordinate systems; the equatorial coordinate system defined by the international celestial reference system (ICRS), and also the Galactic coordinate system. Limiting the coordinate system to these two options is intended to provide a good level of consistency while also accounting for current use of HEALPix. While there has been significant use of HEALPix with the Galactic coordinate system for cosmic microwave background (CMB) data, it is much more common for survey data in other wavebands to be described by equatorial coordinates. ICRS is chosen as the default coordinate system for HiPS with Galactic coordinates as the alternative option.

I don't think introducing a key "ICRS" for gamma-ray astronomy is a good idea if no-one else uses it.
How about we follow what we already define for EVENTS (see http://gamma-astro-data-formats.readthedocs.io/en/latest/general/coordinates.html), i.e. to use ICRS and to use the Galactic system that Astropy uses?
And then put the caveat that Fermi-LAT and some other older data and tools really use FK5 and possibly a slightly different Galactic definition (I had a thread about this with Fermi for their GLON / GLAT in their catalogs and my understanding is that they will change to the Astropy Galactic coordinate definition for future catalogs, again the difference is ~ 10 arcsec I think).

Before we make this decision, let me email the HEALPIX or FITS mailing list and ask what they use / suggest and whether they have plans to standardise this.

Issue #71, opened 2016-11-09
Support for multiple responses (specifically point-like AEFF and EDISP)


cdeil:
The current spec has some support for multiple responses via the `HDU_CLASS` key in the index files.
See http://gamma-astro-data-formats.readthedocs.io/en/latest/data_storage/hdu_index/index.html#hdu-type-and-hdu-class

Specifically for AEFF we do support two versions via the keyword `HDUCLAS2={'EFF_AREA', 'EFF_AREA_RECO'}`
http://gamma-astro-data-formats.readthedocs.io/en/latest/irfs/effective_area/index.html

This is not yet supported in the HDU index file format, there's just one key `HDUCLASS='aeff_2d'` defined, so it's not possible to access both AEFF via the current index file.

In HESS, in addition now we have the use case of wanting to define and access AEFF and EDISP derived from point-like responses (with a `RAD_MAX` cut applied) or full-enclosure responses.
Using point-like responses is very common (in HESS it's used e.g. for all AGN analyses), and we'd like to start using it for FITS analyses.
Note that there is a difference (demonstrated in HESS by @bkhelifi), EDISP is worse for events in the PSF tail, so it's not possible to get the correct EDISP from the full-enclosure responses at a high accuracy level, as another argument that we should support point-like responses in the DL3 spec.

So we would like to  extend the IACT DL3 format spec to support point-like IRFs.

Our main questions are:

- Which AEFF and EDISP header keys should we use?
- How should we support this via the HDU index file?

For the two different AEFF we started to use `HDUCLAS2={'EFF_AREA', 'EFF_AREA_RECO'}`
So we could use `HDUCLAS3={'point-like', 'full-enclosure'}` and add the `RAD_MAX` header key that's already in AEFF to EDISP also.
And then add HDUCLAS2 and HDUCLAS3 columns to the index file spec.
This would satisfy the current use cases for HESS.
It's not very pretty though, because point-like/full-enclosure is not really at a lower or higher level in the hierarchy compared to whether AEFF and EDISP were combined or not. It's just different ways to do responses / analysis.

Another option would be to change the `HDUCLASS` valid keys to `aeff2d_point_reco`, `aeff2d_fullenclosure_reco`, aeff2d_point_true`, `aeff_fullenclosure_true` instead of adding new columns.

[...]

Issue #63, opened 2016-06-28
Consistent column names within IRFs


TarekHC:
Hi all, 

I have a couple of questions regarding the energy naming we use for the IRFs.

Within issue #35 we discussed about true/reconstructed energy, and how to label each of them. We decided to call "ERECO_" and "ENERG_" to reconstructed and true energy respectively. This is inconsistent with the event lists naming convention, but we decided we could live with that (as the same was used in other past experiments).

In the current specs:
- Energy dispersion: "ENERG_" is true energy, so it is ok (although it should be explicitly added). Also we should remove one of the "MATRIX" descriptions currently appearing.
- PSF: Same as before. We should explicitly state that it is True energy
- Background: Here is where I have my main doubts. In IACTs,  we generally use "ERECO" to describe the energy of the cosmics, assuming they are gammas. Shouldn't we then use "ERECO"? Or should we use "ENERG" and add a note there explaining that this is not exactly correct?

I can fix these issues, but I wanted to discuss them first. What do you think would make more sense for the background?

Cheers!

cdeil:
@TarekHC - Thanks for having a detailed look.

Personally I don't care so much about consistent reco / true energy axis names. As you say, in the event list `ENERGY` is reco energy and no-one is in favour of changing this (so far), so full consistency where all reco energy quantities have `RECO` in their name is out the door anyways.

That said, energy axis names should be as simple and consistent throughout, so for sure pull requests to change energy axis names for more consistency or correctness, or alternatively extra comments clarifying which energy axis is reco or true energy are welcome.

For background, as you suspect, it's a reco energy axis. Background is computed and applied directly in reco energy, there is no EDISP associated with the background models. So +1 to clarify this [here](http://gamma-astro-data-formats.readthedocs.io/en/latest/irfs/background/index.html). As I said, I don't care or follow the discussions on energy axis column names much.
Maybe you can make a PR with a suggestion and then we review / decide on it tomorrow in the telcon?

TarekHC:
Hi @cdeil ,

> For background, as you suspect, it's a reco energy axis. Background is computed and applied directly in reco energy, there is no EDISP associated with the background models. So +1 to clarify this here. As I said, I don't care or follow the discussions on energy axis column names much.

Yes, a clarification would be enough. It would be more consistent (at least within the IRFs) to use ERECO, but I don't care that much either.

[...]

PR #79, opened 2017-03-13
Point-like IRFs

TarekHC:
Dear all, 

This is a first draft of the proposal Christoph and I prepared to include the point-like IRF format. The main motivation is to allow within the DL3 format the approach to generate IRFs used by the current generation of IACTs.

You can find an updated version in:
http://www.gae.ucm.es/~thassan/gamma/irfs/index.html

The IRF section was re-organized: IRF components are now described in a dedicated separated section

We can discuss this PR during the open DL3 call and also during the MPIK meeting.


kosack:
is it necessary to have `_full` or `_point` appended to all tables if the `hdu_clas2` already encodes that fact that it is point-like? Isn't that redundant  and makes a "generic" code to load it more difficult?

TarekHC:
> is it necessary to have `_full` or `_point` appended to all tables if the `hdu_clas2` already encodes that fact that it is point-like? Isn't that redundant  and makes a "generic" code to load it more difficult?

I agree. Although, for now, we still do not have a very clear HDU_CLASn hierarchy defined. That is the reason why I was not very cautious with the HDU_CLAS naming (e. g. "bkg_2d", should probably also be divided into "Bkg" and "2D"). We could either wait for a dedicated PR, or I could implement a first approach for the naming convention here, although this might trigger discussion not associated with the new proposed format.

My proposal for HDU_CLAS1 -> HDU_CLAS4 would be (the labels could be changed, focus on the hierarchy):
```
RESPONSE  
        FULL-ENCLOSURE
                AEFF
                        2D
                BKG 
                        2D
                        3D
....               
        POINT-LIKE
                AEFF
                        2D
                BKG 
                        2D
                        3D
....
```

cdeil:
@TarekHC - Thanks!

This PR is the result of #71 and offline discussions.

We will discuss it today at 11 am in the telcon, as well as in the f2f next week.
See https://github.com/open-gamma-ray-astro/2017-03_IACT_DL3_Meeting

People that are involved in IRF production with CTA / HESS / ... or consumption by science tools (Gammapy, ctools) -- please have a look and send comments!
The goal here is to try and agree on something and merge by the end of next week latest.

cc @jknodlseder @jjlk @gernotmaier @bkhelifi @registerrier 

I already made some comments in  #71,  and I will have a closer look at this PR in the coming days.

My feeling at the moment is that:
- the definition of the point-like responses by themselves is pretty straightforward, it's mostly a copy of the full-enclosure responses, with header keys declaring what they are and minor additions (the THETA cut info e.g.)
- there's no concept here how to link EVENTS to IRFs, proposing solutions for that is future work (e.g. updates to the HDU index table or observation definition files), probably based on prototying experience in science tools.
- The HDUCLAS keys will most likely be part of the solution how to locate and interpret IRFs correctly by science tools. Given that we have no concept how they link to EVENTS yet, it's difficult to decide what to put. It's also difficult to decide what piece of info to put at which level of HDUCLAS. My understanding is that so far we didn't really have useful info about HDUCLAS in the spec, and there's no DL3 producers or tools yet, so backwards compatibility is not a concern here.

So overall I'm +1 to make this change & addition to the DL3 spec within the next ~ week as a step in the right direction. It *is* useful because it gives DL3 producers info what to put, and science tool prototyping to use HDUCLAS and to do analyses with both point-like and full-enclosure responses can start. But the understanding is that this will be what we use in the coming month, with the hope that by the end of the year we have a more complete scheme developed (either an evolution of what we're doing here, or just something entirely new that supersedes what we do here).

TarekHC:
As discussed @ Heidelberg, I added a general section introducing the HDUCLASn hierarchy, and added HDUCLASn keywords to all IRFs.

We will probably modify the names of current IRF components format, and even the HDUCLASn labels, but the hierarchy will probably be the one proposed here:

http://www.gae.ucm.es/~thassan/gamma/general/hdu-hierarchy.html

source/general/hdu-hierarchy.rst
```diff
@@ -0,0 +1,87 @@
...
+* HDUCLASS: General identifier of the data format. Recommended value: "OGIP"
+
+* HDUCLAS1: General type of HDU, currently: ``EVENTS``, ``GTI`` or ``RESPONSE``
+
+* HDUCLAS2: In case of ``RESPONSE`` type, refers to the IRF components stored within the HDU: ``EFF_AREA``, ``BKG``, ``EDISP`` or ``RPSF``
+
+* HDUCLAS3: In case of ``RESPONSE`` type, refers to the way the IRF component was produced (``POINT-LIKE`` or ``FULL-ENCLOSURE``)
+
+* HDUCLAS4: In case of ``RESPONSE`` type, refers to the name of the specific format
+
```
cdeil: The description of what to put for HDUDOC could be central, putting the Github URL.
For HDUVERS I would suggest to put "0.2" or "1.0" for now for all formats, and then on the general page describe that they can evolve differently, that it's not coupled to the global version from this spec.

source/general/hdu-hierarchy.rst
```diff
@@ -0,0 +1,87 @@
+.. include:: ../references.txt
+
+.. _hdu-hierarchy:
+
+HDU hierarchy
```
TarekHC: Fixed

source/general/hdu-hierarchy.rst
TarekHC: Added a section describing briefly the different types of HDUs in the hduclass section.
Added a reference to that section in the definition of HDUCLASS keywords.
Currently we just define HDUDOC and HDUVERS in this section. I would try to de-correlate this versioning discussion with this PR... Or we will never merge.

source/general/hdu-hierarchy.rst
```diff
@@ -0,0 +1,87 @@
...
+* HDUCLAS4: In case of ``RESPONSE`` type, refers to the name of the specific format
+
++------------+-----------+------------+-----------------+--------------+
+|  HDUCLASS  |  HDUCLAS1 |  HDUCLAS2  |  HDUCLAS3       |   HDUCLAS4   |
++============+===========+============+=================+==============+
+|    OGIP    |    GTI    |            |                 |              |
++------------+-----------+------------+-----------------+--------------+
+|            |  EVENTS   |            |                 |              |
++------------+-----------+------------+-----------------+--------------+
+|            |  RESPONSE | EFF_AREA   |   POINT-LIKE    |   aeff_2d    |
```
TarekHC: Argh, that was already corrected, and I did not push the commits... Now it should be correct.

TarekHC:
Hi @cdeil ,

Thank you for the review!

I believe I incorporated all your suggestions. There were some that I had fixed locally during the meeting but forgot to push... (yes, I know, it's annoying... sorry about that!)

I also reformatted the point-like section as you suggested. As you said, I have the feeling that when somebody is taking a look at the specifications of a format, they really just want to see the whole thing together, without needing to go through different sections to understand the little differences. But of course for us it would be a pain to maintain both sections, so let me know if you like the new format.

I also added "True" or "Reconstructed" to all energy axes (following discussion in #63 ). The labelling is still terribly inconsistent (background ``ENERGY`` is reconstructed, while in the effective area is ``ENERGY`` refers to true energy and ``ERECO`` to the reconstructed one... a mess!).

You can find the updated specs online at:
http://www.gae.ucm.es/~thassan/gamma/index.html

cdeil:
@TarekHC - Thanks!

I see that now e.g. for EDISP the page
http://www.gae.ucm.es/~thassan/gamma/irfs/full_enclosure/energy_dispersion/index.html
is separate from the description of EDISP at
http://www.gae.ucm.es/~thassan/gamma/irfs/irf_components/index.html#energy-dispersion

@TarekHC - Wouldn't it be better to keep them together?

If you want to keep as-is, please add links from the pages under `irfs/full_enclosure` to the sections in `irfs/irf_components`. (currently there are statements like "Energy migration axis (defined above)", but it's no longer "above" but on a separate page now).

TarekHC:
I think that is probably just a matter of taste. 

I would go for leaving as it is (and adding references, as you suggest). I think separating the IRF component "formal" definitions from the description of the format(s) (technical and more "pragmatic") makes sense. Although others may not like it, so I'm totally open to both possibilities.

I will add the references now, and if anyone has a strong opinion, then I will change it back.

TarekHC:
I think that is probably just a matter of taste. 

I would go for leaving as it is (and adding references, as you suggest). I think separating the IRF component "formal" definitions from the description of the format(s) (technical and more "pragmatic") makes sense. Although others may not like it, so I'm totally open to both possibilities.

I added the references, and if anyone has a strong opinion, then I will change it back.

cdeil:
@TarekHC - Agreed that structuring of IRF description is a matter of taste and both are valid ways to put it. Merging as-is now. Thanks for all your work on this pull request and patience with the review!

PR #82, opened 2017-04-23
Rename results section to spectra

cdeil:
This pull request renames the "high-level results" section to "spectra".

For CTA they have the organisation where DL3 is EVENTS + IRF and then maps and cubes and spectra are DL4. I feel the distinction what is "high level" within the specs we have here and what isn't is pretty arbitrary, so I think just getting rid of the "high level" label altogether is better.

@woodmd - OK?


woodmd:
I agree that high-level isn't a particularly useful distinction so these changes are fine with me.

PR #83, opened 2017-04-24
Use "full" less where not needed (to be simpler)

cdeil:
This PR reverts some of the additions of "full" that @TarekHC did in #79 .

Changing Sphinx labels breaks links to the spec from outside, e.g. from the Gammapy docs:
https://travis-ci.org/gammapy/gammapy/jobs/224927389#L2002

The "full" was added in three places:
- Sphinx labels
- Section titles
- As values in the HDU index tables.

I feel that the way the "full" was added isn't helpful, i.e. no corresponding "point" was added to the spec and then one might just as well leave the distinction up to the HDUCLAS3 keyword.

Gammapy is the only software package affected by this change, so I'm merging this now.
I'll do a few follow-up pull requests in the next days to clean up and go towards the first stable version. I'll keep the PRs small so that the diffs are easy to review.

@TarekHC - If you disagree with any of my edits, let me know and we can talk about it on Skype.

### DIFF
--- a/source/data_storage/hdu_index/index.rst
+++ b/source/data_storage/hdu_index/index.rst
@@ -65,26 +65,18 @@ Valid ``HDU_TYPE`` values (others optional):
 Valid ``HDU_CLASS`` values:
 
 * ``events`` - see format spec: :ref:`iact-events`
-* ``gti`` - see format spec: TODO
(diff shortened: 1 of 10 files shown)


TarekHC:
Yes, I agree. Did not check if you fixed the labelling everywhere, but at least the changes you made make sense.

PR #85, opened 2017-04-24
Removing ERECO column, and reducing AEFF vs ERECO section

TarekHC:
Following issue #63:
* Reduced significantly "Effective Area vs reconstructed energy" section
* Updated "create_example_file.py" script (also including HDUCLAS hierarchy)
* Created example fits with "create_example_file.py"


cdeil:
@TarekHC - Thanks!

Looks good to me. Merging this now.

As mentioned in https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/issues/63#issuecomment-296102281, no open-source tools were supporting the previous `ERECO` column name yet, so I think it's OK to put this in without discussion.

If you disagree with this change, please comment here.

TarekHC:
@cdeil , this was never merged (although you did say to merge it now).

I would merge it myself, but I also added the example bkg_2d and 3d scripts/files. It is not very organized but shall we merge it directly with both additions?

cdeil:
Merging it now.
No, really, this time I mean it!
:-)

Thanks!

Issue #86, opened 2017-04-24
how to add size of energy bin to differential flux points


GernotMaier:
Is there a way to give the min/max of an energy bin in a differential flux spectrum? 

Some spectra in the literature are given with uneven binning, and it would be useful to store this information somehow.

cdeil:
You can add `e_min` and `e_max` to denote the reco energy bin in which the differential flux measurement was made. At least that's what I did for gamma-cat. One example: https://github.com/gammapy/gamma-cat/blob/master/input/data/2005/2005A%2526A...437L...7A/tev-000039-sed.ecsv

Format spec: http://gamma-astro-data-formats.readthedocs.io/en/latest/spectra/flux_points/index.html#energy-columns

@woodmd - I'm assigning this issue to you. Either close, or if you have time, you could also add this to one of the example files in the spec and explicitly state that using `e_min` and `e_max` for differential spectra is allowed?

woodmd:
The intention was that `e_min` and `e_max` could apply to both integral and differential representations of the flux measurement.  I can add some text to the spec to make this more explicit.

cdeil:
@woodmd - Thanks!

The way I'm using `e_min`, `e_max in Gammapy is as follows:
- For differential spectral points, it's the **reco energy band** in which the spectral point was measured
- For integral spectral points, the assumption is that that measurement has taken care of energy dispersion somehow and the quoted integral flux is for a **true energy band**.

I don't know if this way to fill and interpret `e_min`, `e_max`, with different semantics for differential and integral spectral points, is what you intended or use it yourself. But @woodmd, it's a non-obvious point that probably others will wonder about too as they do measurements where energy dispersion has a big effect (e.g. pulsar spectra from TeV instruments), so it's good if there's a note on that point in the spec.

woodmd:
I've opened #87 which includes some text to clarify that `e_min` and `e_max` can apply to differential quantities.

Regarding reco vs. true energy my intention was that SEDs would be used only for measurements in true energy since this is generally what you are interested in when you are comparing to a measurement from another instrument or a theoretical model.  If you distribute measurements in reconstructed energy it's not clear to me how one would use those without also providing additional information about the instrument response (e.g. an energy dispersion matrix).  I also don't see any particular reason to make a distinction between differential and integral measurements with respect to the influence of energy dispersion.  

[...]

PR #87, opened 2017-04-25
SED spec updates

woodmd:
This PR makes a few minor changes to the SED spec:

* Drops `ph` unit everywhere as requested in #81.
* Clarifies that `e_min` and `e_max` can apply to differential representations (see #86).
* Cleanup example files and add python script for regenerating these files automatically.


cdeil:
Looks good to me.

PR #90, opened 2017-07-12
Update recommended value for HDUCLASS in hduclass.rst

lmohrmann:
This request updates the recommended value for the HDUCLASS header keyword from "CTA" to "GADF" as already introduced e.g. [here](http://gamma-astro-data-formats.readthedocs.io/en/latest/irfs/full_enclosure/aeff/index.html).

### DIFF
--- a/source/general/hduclass.rst
+++ b/source/general/hduclass.rst
@@ -28,7 +28,7 @@ The different HDUs defined in the current specifications are listed here:
 
 The current HDU class scheme used is the following:
 
-* ``HDUCLASS`` : General identifier of the data format. Recommended value: "CTA"
+* ``HDUCLASS`` : General identifier of the data format. Recommended value: "GADF" (for "gamma-astro-data-formats")
 * ``HDUDOC`` : Link to the DL3 specifications documentation
 * ``HDUVERS`` : Version of the DL3 specification format
 * ``HDUCLAS1`` : General type of HDU, currently: ``EVENTS``, ``GTI`` or ``RESPONSE``


cdeil:
There was quite some discussion here on what value to put for this key in the context of CTA:
https://github.com/cta-observatory/ctapipe/issues/308#issuecomment-296487892

It's mostly me suggesting to put "GADF" for "gamma-astro-data-formats" and @jknodlseder  saying we should use OGIP.

I think when we ( @TarekHC , @cboisson, @bkhelifi ) discussed this at the DL3 f2f meeting, we agreed to put "GADF" and did that everywhere in the spec, and this one remaining mention of "CTA" is an oversight?

Just for context: @lmohrmann is updating the HESS FITS exporter, and we want to finalise this now that the HESS FITS test data release is approved, and produce files that are up to this spec.
To me "CTA" (what's mentioned there now) isn't a great choice, since this is something that existing IACTs also want to use and it's not been approved or decided on in CTA, and other places in the spec currently mention "GADF" already.

I would suggest to merge this now, and then if / when in CTA there's a DL3 group that can decide something in the coming years, the discussion on whether to put "OGIP" or "CTA" instead can continue.

TarekHC:
> I would suggest to merge this now, and then if / when in CTA there's a DL3 group that can decide something in the coming years, the discussion on whether to put "OGIP" or "CTA" instead can continue.

I'm ok with this change. In fact, in every IRF described, we suggest to use the "GADF", so probably this line was simply missing to be updated.

So I would not re-open the discussion and directly merge it.

Issue #91, opened 2017-07-14
Inconsistent specifications for HDUCLAS2 keyword in EVENTS extension


lmohrmann:
The specifications for the HDUCLAS2 header keyword in the EVENTS extension are inconsistent: HDUCLAS2 is listed as a mandatory (!) header keyword [here](http://gamma-astro-data-formats.readthedocs.io/en/latest/events/events.html), but not specified at all for the EVENTS extension [here](http://gamma-astro-data-formats.readthedocs.io/en/latest/general/hduclass.html).

cdeil:
The `HDUCLAS2` for `EVENTS` was introduced by @jknodlseder [here](https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/commit/68c9aaf6cf71ad74eea11552e98158e766433d96#diff-ab2e58774ff2e0930caa191b5c3a8ecfR119) and then that bullet was slightly modified by @TarekHC (I think, not sure if that is your commit) [here](https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/commit/a8c089d30989f2d2815a834a1e3b76ea9091a4a6#diff-ab2e58774ff2e0930caa191b5c3a8ecfR111).

This key is mentioned here in this OGIP recommendation, and they don't make the distinction optional / required like we do:
https://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/ofwg_recomm/hduclas.html

I don't think science tools currently read and use this keyword, so my suggestion would be to change it to optional and not mention it in the summary table. But I don't care much either way how this detail is changed.

@jknodlseder @TarekHC - Thoughts?

TarekHC:
> I don't think science tools currently read and use this keyword, so my suggestion would be to change it to optional and not mention it in the summary table. But I don't care much either way how this detail is changed.

I'm not sure if I agree here. Data formats should be as stable as possible in time (so we may not used them now, but maybe in the future), and having some extra keywords formally describing each table class makes sense. I would personally make them mandatory, even if for the moment they are not used (but may help astronomers understand HDUs a bit better...).

lmohrmann:
Then the table on [this page](http://gamma-astro-data-formats.readthedocs.io/en/latest/general/hduclass.html) should maybe be updated? Are there other suggested/possible values for HDUCLAS2 other than ACCEPTED?

cdeil:
[...]

PR #92, opened 2017-07-19
Updates to EVENTS extension

lmohrmann:
HDU header keywords may not be longer than 8 characters. I therefore propose to change EVENT_CLASS to EV_CLASS.

### DIFF
--- a/source/events/events.rst
+++ b/source/events/events.rst
@@ -107,8 +107,6 @@ Mandatory header keywords
     * Version of the format (e.g. '1.0.0'). See :ref:`hduclass`.
 * ``HDUCLAS1`` type: string
     * Primary extension class (option: 'EVENTS'). See :ref:`hduclass`.
-* ``HDUCLAS2`` type: string
-    * Secondary extension class (option: 'ACCEPTED'). See :ref:`hduclass`.
 * ``ORIGIN`` type: string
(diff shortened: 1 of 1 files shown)


cdeil:
Rant: This is a stupid limitation of FITS, forcing us to have cryptic names and inconsistent column names and header keys.

I see that Fermi-LAT has an `EVENT_CLASS` column:
https://fermi.gsfc.nasa.gov/ssc/data/analysis/documentation/Cicerone/Cicerone_Data/LAT_Data_Columns.html#PhotonFile
They don't have it as a header key (at least in the one file I checked).

As far as I can see the OGIP spec doesn't mention such a key:
https://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/events/ogip_94_003/ogip_94_003.html

As far as I know, the current science tools don't use that key for anything.

So I'm OK with the change proposed here, to make our EVENTS spec a valid FITS file.

@jknodlseder and anyone else interested - Thoughts?

lmohrmann:
Following the discussion in #91, I changed to HDUCLAS2 header keyword in the EVENTS extension to be optional.

cdeil:
No further comments / complaints in the past 23 days. Merging now.
@lmohrmann - Thanks!

Issue #62, opened 2016-06-17
How to handle energy dependent deadtime/livetime?


jknodlseder:
The deadtime (or livetime) for CTA is in principle energy dependent (as different telescopes and cameras will have different deadtimes). The question is how to handle that. Fold the deadtime into the IRF? Of having explicitly an energy dependent deadtime? This is even worse if there are different camera types on the same site, and events seen with one camera type will have a deadtime different to events seen by other camera types.

cdeil:
My understanding is that science results always depend on the product `aeff * obs_time * (1 - dead_time_fraction)`, so whether to store the dead time fraction separately, or multiply it in DL3 with `aeff` or `obs_time` is a matter of taste how one wants to structure the book-keeping.

If CTA has complicated dead time fractions, it might be the easiest to produce DL3 where `dead_time_fraction=0` and the dead-times have been applied to `aeff`.

What is CTA doing at the moment concerning dead time in prod2 and prod3 IRFs?

kosack:
Does the per-telescope deadtime really matter? Isn't it the overall deadtime as measured by the SWAT (central trigger software) the more important quantity, which is a sort of "average" or "effective" deadtime for the system.     It's similar to other quantities that we may consider as averages like the trigger pattern.  Of course one could imagine if we really decide to split IRFs by trigger pattern (probably not easy due to the number involved), then it may make sense to talk about dead-times per effective multiplicity group (section of subarray triggered), etc.

cdeil:
> Does the per-telescope deadtime really matter? 

No, at DL3 per-telescope quantities don't matter.
The data model is EVENTS and IRFs for sub-arrays.

Everything per-telescope should happen at lower levels as part of the DL3 production.

One thing that is missing in the current DL3 spec is "event types", sub-partitions of the events by quality. These could be split by trigger pattern or other things. Each has their own effective area and deadtime fraction.

jknodlseder:
I think nothing is done in Prod2 or Prod3 with the deadtime, Gernot may confirm.

As I understand so far from discussions with Karl and Gernot, deadtime depends on trigger pattern (which cameras triggered, hence on energy) and time (for example rising moon). On VERITAS the deadtime is computed on a minute time scale.

[...]

PR #97, opened 2017-10-23
Clarify background rate per livetime or obstime

cdeil:
The specs for background rate aren't clear if it's per livetime or per observation time:
http://gamma-astro-data-formats.readthedocs.io/en/latest/irfs/full_enclosure/bkg/index.html

This needs to be clear for science tools to correctly compute predicted background counts.

@jknodlseder - How do you interpret the values in ctools?

@gernotmaier - How do you fill background in the CTA IRF FITS exporters?

### DIFF
--- a/source/irfs/full_enclosure/bkg/index.rst
+++ b/source/irfs/full_enclosure/bkg/index.rst
@@ -31,10 +31,6 @@ Required columns:
 * ``BKG`` -- ndim: 2, unit: s^-1 MeV^-1 sr^-1
     * Absolute post-select background rate
       (expected background per time, energy and solid angle).
-    * Note that this is not a "flux" or "surface brightness".
-      This is already a count rate, it doesn't need to be multiplied with
-      effective area to obtain predicted counts, like gamma-ray flux and
-      surface brightness models do.
 
 Recommended axis order: ``ENERGY``, ``THETA``
 
@@ -70,10 +66,6 @@ Required columns:
 * ``BKG`` -- ndim: 3, unit: s^-1 MeV^-1 sr^-1
     * Absolute post-select background rate
       (expected background per time, energy and solid angle).
-    * Note that this is not a "flux" or "surface brightness".
-      This is already a count rate, it doesn't need to be multiplied with
-      effective area to obtain predicted counts, like gamma-ray flux and
-      surface brightness models do.
 
 Recommended axis order: ``ENERGY``, ``DETX``, ``DETY``
 
@@ -91,3 +83,14 @@ declare the type of HDU:
 * ``HDUCLAS4`` = 'BKG_3D'
 
 Example data file: :download:`here <./bkg_3d_full_example.fits>`.
+
+
+Notes
+-----
+
+The background rate is not a "flux" or "surface brightness".
+It is already a count rate, it doesn't need to be multiplied with
+effective area to obtain predicted counts, like gamma-ray flux and
+surface brightness models do.
+The rate is computed per observation time (without any dead time correction,
+don't use livetime when computing or using the background rate).


cdeil:
I've added one clarifying sentence here that background rate is per observation time, not livetime in ecf93f5 . I've put it in a "notes" section at the end, like we already have on other pages, such as e.g. the EVENTS page.

@TarekHC - Please review. Let me know if there's something that should be improved, or merge if it looks good to you.

@lmohrmann - You're working on HESS background rate models - you're filling the rates per observation time (not livetime), right?

@jknodlseder - I think ctools already uses background rate like this, so nothing to do there, no?

(I think all code in Gammapy to produce and use background rate models, like [here](http://docs.gammapy.org/en/latest/api/gammapy.cube.make_background_cube.html) added today, is already using it as a rate per obs time, so nothing left to adjust there)

source/irfs/full_enclosure/bkg/index.rst
```diff
@@ -91,3 +83,14 @@ declare the type of HDU:
...
+
+
+Notes
+-----
+
+The background rate is not a "flux" or "surface brightness".
+It is already a count rate, it doesn't need to be multiplied with
+effective area to obtain predicted counts, like gamma-ray flux and
+surface brightness models do.
+The rate is computed per observation time (without any dead time correction,
```
TarekHC: dead-time correction

TarekHC:
Everything looks ok to me. Merged.

jknodlseder:
ctools and GammaLib interpret background rate per livetime

jknodlseder:
GammaLib / ctools has the convention that counts = rate * livetime, hence all rates are per livetime.

cdeil:
@GernotMaier - You could consider adding one sentence "background rate is given per observation time, not livetime" to the CTA IRF description document you wrote, to make it clear what is given.

@jknodlseder - Would you consider changing Gammalib / ctools to have background rate per observation time instead of per livetime?

---

Given that CTA dead time fraction will be very small (~1%) and at the moment is not considered at all in the simulations, it doesn't really matter for results (and in addition most analyses have a free scale parameter on the background, i.e. results don't pend on the absolute rate in the background IRF model). So making any changes here isn't high priority, but someone will have to pin down the IACT DL3 details at some point. In the end it's just a matter of taste where the book-keeping for deadtime fraction is done, and for now we have decided to just follow in the spec what DL3 producers already do (i.e. background rate per observation time). @TarekHC already linked to #62 above where deadtime fraction is discussed in the context of gammas / effective area. I'm not sure, but again I think actually what we have in the open spec is exactly what DL3 producers already do, and might be sufficient for CTA (if they want to absorb dead time fraction in effective area, they can just set DEADC header key to 0).

GernotMaier:
I've updated the document - there are a few more minor things to do before sending out a new version. For those with access, the most recent version can always be found on the CTA svn:

https://forge.in2p3.fr/projects/cta_analysis-and-simulations/repository/changes/DOC/InternalReports/IRFReports/trunk/cta-aswg-IRFreport.pdf

jknodlseder:
I would not like to change the definition of counts and rate as this would be inconsistent between source and background counts.

cdeil:
I hadn't realised that in Gammapy we had introduced a background computation as `counts = rate * livetime`. Following the discussion above, we have changed to `counts = rate * obstime` in https://github.com/gammapy/gammapy/pull/1861 .

Overall by now I like the proposal some people made in #62 to get rid of DEADC at the DL3 level, and just define AEFF and BKG rate such that it contains dead-time effects, and science tools only need to use obstime. Especially if we use GTI tables, it's simpler if one doesn't also need to query the EVENTS header to get a DEADC key to get a livetime and compute and exposure or predicted background.

@jknodlseder - For ctools are you willing to change to this definition? Or do you prefer to keep as-is, and wait for an official CTA decision on this before changing the software?

PR #98, opened 2017-10-25
Updates to HPX and WCS Map Formats

woodmd:
This PR brings the format spec for HPX and WCS maps up-to-date with some of the recent work we've being doing in `gammapy.maps`.  This fleshes out the section on WCS maps and describes the formats used in `gammapy.maps` for sparse and non-regular/multi-resolution maps (mostly inspired by the HPX spec).  I also took the opportunity to improve the organization a bit by moving the documentation for the BANDS table into the main skymap page (since it's common to both WCS and HPX maps).  Here are the two main changes to the HEALPix spec:
* Add an `AXCOLS{X}` header keyword to the BANDS table which specifies the mapping of columns to a given dimension of the map.  This change is also propagated to the sample files.
* Add a `LOCAL` indexing scheme.  This is similar to the `EXPLICIT` scheme but uses uses a local pixel index that supports different pixelization in each band.  The `SPARSE` scheme also supports multi-resolution geometries but there may be some use cases where this format is preferable.

There are probably a few parts of the spec that could be improved.  The description of how multi-resolution WCS maps work could probably stand be expanded.  We're also lacking sample WCS files.  

One question I wanted to raise with this PR is how people feel about the current mechanism for discovering the axis column names in the BANDS table.  The current spec allows the axis columns to have any name (e.g. `ENERGY`, `TIME`, etc.) and uses the `AXCOLS` keyword to specify the mapping of these columns to an axis of the map.  This was partially motivated to make the spec backward compatible with the existing convention for energy cubes where these columns are called `ENERGY` or `E_MIN`, and `E_MAX`.  

The alternative approach would be to fix the column names (`AX0_MIN`, `AX0_MAX`) and use a header keyword to specify the name of the column (e.g. `AXNAME0 = ENERGY` or `AXIS0 = ENERGY`).  Of course with this convention we would need to make an exception in the format to support existing cube files (e.g. if HDU is called `EBOUNDS` then use `E_MIN` and `E_MAX`) but that's already the case in the existing spec as well.  I kind of prefer this latter approach since it's closer to what we do for the image data where column names are fixed and don't depend on the type of quantity in the file.  I think it probably also simplifies the logic one would need to use to validate a file.


cdeil:
I've put a rendered HTML version of the spec from this PR here to make review easier:
https://www.mpi-hd.mpg.de/personalhomes/deil/public/gamma-astro-data-formats/pr-98/

@jknodlseder - Are you interested in these maps data formats developed by @woodmd that support event types (via the general concept of BANDS), as well as WCS and HPX maps and multi-resolution and sparse maps for Gammalib / ctools? If yes, could you please review the formats and give feedback?

@tburnett - Any thoughts / feedback on this, given your experience with`pointlike`?

I think @adonath or @facero might also be interested / have feedback concerning these maps formats - if you have time to review, please do and leave comments here.

---

@woodmd - My experience is that it's hard to get good feedback on PRs, and feedback will only come later when people try to use this format or implement it. So if you don't get good feedback here, I would suggest you just go ahead and merge next week using whatever scheme / formats you prefer, and then we advertise this more widely in the coming weeks and try to get feedback.

---

Concerning content organisation - I think it's great to have the description of BANDS split out if it's used by WCS and HPX. But things should be re-arranged so that "WCS maps" and "HEALPix maps" appears as sub-points of "Sky maps" in the navigation / outline, no?

This is how it appears at the moment:
screen shot 2017-10-29 at 16 07 25

My other comment would be that example files are very important - if you can add some more here (https://www.mpi-hd.mpg.de/personalhomes/deil/public/gamma-astro-data-formats/pr-98/skymaps/wcs/index.html#sample-files) with limited effort, please do!

> One question I wanted to raise with this PR is how people feel about the current mechanism for discovering the axis column names in the BANDS table.

I don't understand the pros / cons of the two approaches you outline. I think both require a tool that knows about the serialisation format, i.e. in either case it's not possible to open up such a map with a generic sky map tool  like DS9 or Aladin and get a good representation of the image. Given that there is no astronomy standard / tool support for these formats, I think the serialisation format doesn't matter so much any more. There are still concerns of course like simplicity, extensibility, as few special cases as possible, yet maybe supporting existing files from Fermi-LAT, but to be honest I don't have enough knowledge on this topic or time to think through the different cases, so probably won't be able to give good advice here. Maybe if you post example files here I could browse them a bit and see if the format / information given is intuitive / understandable and try to give feedback. But otherwise @woodmd - I would propose that you just go ahead with the option you prefer, and then we'll try to advertise this and get feedback from real-world usage in the coming months.

cdeil:
I've made a notebook that shows the resulting FITS serialisation of some of the WCS map examples from the Gammapy docs: https://nbviewer.jupyter.org/gist/cdeil/b0c8f2cae94dc1cf4189f84dd2376a79

Here is the multi-resolution case, with an energy-dependent pixel size
https://nbviewer.jupyter.org/gist/cdeil/b0c8f2cae94dc1cf4189f84dd2376a79#3-dim-map-(multi-resolution)

I'm a bit confused about that case. So there's a data array of size `(99, 99, 12)`, i.e. the same number of pixels in each energy band, and they cover a different part of the sky? Is this useful?

I would have expected WCS-based multi-resolution to instead have roughly or exactly the same spatial coverage in each energy band (e.g. by going in steps of 2 in resolution), which would require a completely different FITS format (e.g. a BANDS table listing the energy bands, and then one IMAGEHDU with a 2-dim array it's own WCS per energy band). I guess this isn't supported by the current spec?

@woodmd - Maybe you could make an example file for the ~ 3 most common use cases supported by the WCS maps formats you're adding here, and then link to a notebook like I did that shows the FITS serialisation in the formats defined here?

woodmd:
> But things should be re-arranged so that "WCS maps" and "HEALPix maps" appears as sub-points of "Sky maps" in the navigation / outline, no?

I agree but I wasn't able to figure out how to get them to appear here.  I guess I need to create a hidden table of contents in the skymap page?

> My other comment would be that example files are very important 

Ok I can add these.  It's quite easy since it's already the format understood by `gammapy.maps`.

>  I would propose that you just go ahead with the option you prefer, and then we'll try to advertise this and get feedback from real-world usage in the coming months.

Ok that's fine.  I think approaches would work so it's more a matter of taste.  

> I guess this isn't supported by the current spec?

You can do this but you would need to define both the pixel size and number of pixels per band.  So taking your example you would do:
```
npix = (10.0/binsz).astype(int)
m_wcs = MapBase.create(binsz=binsz, map_type='wcs', skydir=position, width=10.0,
                       axes=[energy_axis], npix=npix)
```
This ensures that all bands have approximately the same ROI diameter.  

Note that ROIs with different spatial coverage in each band is potentially useful if you want to ensure that the PSF is fully encompassed by the ROI at low energies but not too large at high energies.  In the LAT catalog analysis we currently employ a strategy like this using joint likelihood between analysis components with different pixel and ROI size.  

> Maybe you could make an example file for the ~ 3 most common use cases supported by the WCS maps formats you're adding here, and then link to a notebook like I did that shows the FITS serialisation in the formats defined here?

OK will do.

jknodlseder:
I certainly would be interested to check the proposed format

woodmd:
I've now added sample WCS files as requested by @cdeil and made a few other miscellaneous improvements:
* Removed `AXCOLS` as header keyword from MAP tables.
* Switched to 1-based index for `AXCOLS` header keyword.  
* Add text to clarify that CHANNEL to band index mapping follows column-major ordering convention.

Since I didn't receive any feedback so far I'm going to go ahead and merge these changes now.  If anyone has additional comments or feedback on the current spec feel free to open a new issue.

cdeil:
@woodmd - Thank you! It's great to have a spec and example files for software developers and advanced users!

All - you can find the latest version of these specs here: http://gamma-astro-data-formats.readthedocs.io/en/latest/skymaps/ -- please review!

Is this section still to be written? Or did you mean to put it above the `FGST` and `GALPROP` format description above?
http://gamma-astro-data-formats.readthedocs.io/en/latest/skymaps/healpix/index.html#non-standard-healpix-conventions

Is there a reason to prefer 1-based over 0-based indexing for AXCOLS or is it just a choice you made, but really there aren't any pros or cons either way?

tburnett:
Yes, I'm impressed with this effort, my thanks to Matthew as well.


The SPARSE option is  based on a format that I've been using for 10 years. There are two energy-dependent factors important for representing Fermi data: the PSF size falls roughly like 1/E until 10 GeV or so, and typical spectra fall like 1/E**2. For this format, we chose pixel sizes to be a fraction of the PSF. The two factors mean that the pixel densities vary from from 1e3 counts/pixel for our full dataset at low energies to 1e-6 at the highest energies. This is with nside varying from 86 to 8192.  The file size for the 8-year dataset that we plan to use for the upcoming 4FGL catalog occupies 143 MB.

It contains 15M pixels, and 220M photons.


I can now convert my FITS files to this SPARSE format, will be using it in the future. Also, I expect to use gammapy to create them in the future.


--Toby

________________________________

woodmd:
> Is this section still to be written? Or did you mean to put it above the FGST and GALPROP format description above?

Yea this is still on my to-do list.  These aren't formally part of the spec itself but it's probably good to have so that we can ensure compatibility with files written by these codes.

> Is there a reason to prefer 1-based over 0-based indexing for AXCOLS or is it just a choice you made, but really there aren't any pros or cons either way?

I felt this was preferable since it matches to the convention of WCS keywords which also use 1-based indexing (`CTYPE1`,`CDELT1`, etc.).

> This is with nside varying from 86 to 8192. 

We probably need to have some discussion about the extent to which we want to support nside values that are not a power of 2.  The spec itself is totally agnostic to this but there are some complexities that come with supporting both cases in software.  In `gammapy.maps` we currently assume that nside is always a power of 2 but I don't think it would be a huge effort to support non-power of 2 nside values.  The primary change would be to disallow certain operations like changing map resolution.

@tburnett I just checked with the latest version of healpy and it appears that non-power of 2 nside values are supported:
```
In [1]: import healpy as hp

In [2]: hp.__version__
Out[2]: '1.11.0'

In [3]: hp.ang2pix([7,8], 0.5,0.5)
Out[3]: array([25, 41])
```

tburnett:
About non-power of 2 nside:


healpy does not support query_disc with nside not a power of 2. This is crucial for me to extract the subset of data pixels within a cone.

I often use aladin for display, which enforces the restriction. Since it is easy to sample to a larger nside, I just convert maps if I want to make such a display.

The primary reason for the nside choice is storage efficiency: from one level to the next is a factor of 4.

________________________________

woodmd:
> healpy does not support query_disc with nside not a power of 2. This is crucial for me to extract the subset of data pixels within a cone.

Ok that would also be a problem for `gammapy.maps` since we also rely on that function.  @cdeil do you know the extent to which `astropy-healpix` will support nside not a power of 2?

PR #81, opened 2017-10-29
Remove "ph" for "photon" from units

cdeil:
I would like to propose to consistently remove "ph" for "photon" in unit strings.

I.e. write e.g. "s-1" instead of "ph s-1".

The motivation for this change is that putting "ph" or not results in incompatible units:
```
>>> import astropy.units as u
>>> u.Unit('ph s-1') == u.Unit('s-1')
False
```
forcing packages and tools that want to use and check units to standardise on either always having "ph" or not. E.g. in Gammapy we had to add this to make reading and working with the example SEDs given in this spec repo work:
https://github.com/gammapy/gammapy/blob/998714db4d799521f9d71d7880abe7190ea995cb/gammapy/spectrum/flux_point.py#L76

Currently some places in the spec don't have "ph" (e.g. the rates in the background models) and the SED example files and description do have it:
https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/search?utf8=%E2%9C%93&q=ph&type=

@woodmd (and maybe also @adonath or @joleroi) - Thoughts?

Assuming yes, note that the "ph" is also in the example SED FITS files in this repo like https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/blob/master/source/results/flux_points/flux_points.fits which I can't easily update. Ideally we would switch to example files in the spec that are always generated by scripts (and in many cases could have simple dummy values or zeros) to make this spec more maintainable (even if the example files then don't contain nice data).


cdeil:
@woodmd - I think this discussion concluded and we just forgot to close the issue?

I found one last place where the example files contain "ph":
https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/blob/83cd4d2c9487aae51956a6d2ef99edd4c5b9e87b/source/spectra/flux_points/binlike.ecsv#L9

Fixed in a commit  0333d79 which I attach here. @woodmd - please review.

cdeil:
I'm merging this small update to the flux point example files in now. I think it's completely uncontroversial, just a small cleanup. See my last comment.

Issue #95, opened 2017-10-19
Format of TSTART and TSTOP in obs-index table not defined consistently


lmohrmann:
The format of the fields TSTART and TSTOP is not defined consistently. On [this page](http://gamma-astro-data-formats.readthedocs.io/en/latest/general/time.html), it says the values are stored in MJD, whereas [this page](http://gamma-astro-data-formats.readthedocs.io/en/latest/data_storage/obs_index/index.html#obs-index) says they are stored in MET. Which one do we want?

Note that TSTART and TSTOP are also keywords of the EVENTS extension, where they are stored in MET.

cdeil:
@lmohrmann - Thanks for filing the issue!

I think we struggled with adopting the FITS time standard with the reference time header keywords as used in EVENTS for OBSINDEX because we said it should be more like a "runlist" that is human-readable and that can be stored in CSV files that don't support a header and that time format.

But looking back, I think that was a mistake, because it causes a lot of complexity with time-related information, and we should just require the same time format that we already use for EVENTS.

My understanding is that backward compatibility isn't a bit concern yet.
The first observation index for CTA was made recently, and uses the FITS way to define times already (I hope):
https://github.com/gammasky/cta-dc/blob/e6cdb0567340d23fa14fe9ae400d3bcdb7899281/data/cta_1dc_make_data_index_files.py#L215
MAGIC is just starting to use this, and the main use comes from HESS and Gammapy, and we should really aim to clean this up before the first HESS data release, to avoid having a mess for the coming years.

@lmohrmann - Can you make a pull request to improve this aspect (or even anything else that you notice) concerning the obs index table?

PR #103, opened 2018-01-18
Improve and de-duplicate time specifications

cdeil:
This pull request improves and avoids duplication in the specifications of times.

The only change in format is for the observation index files, where "DATE_OBS" with an underscore was used, whereas e.g. for the events "DATE-OBS" with a minus was used. My understanding is that this was a mistake, the FITS standard and also OGIP documents I found use "DATE-OBS" with a minus. I don't think that change is controversial, we already said in #95 that we want to use the same format for the observation index file that we use for EVENTS, and also almost no-one except us likes or wants to use the obs index tables anyways (see #7)

Another place where I saw "DATE_OBS" with a minus is in the CTA 1DC files produced by ctobssim, which I've reported  at https://forge.in2p3.fr/boards/243/topics/2427 (cc @jknodlseder).

Apart from that small change / mistakes, the big change here in the spec is a simpler and rewritten `general/time.rst` page, and to reference that from other specs that contain times. Currently we had duplicated and slightly inconsistent and wrong descriptions four times (EVENTS, GTI, POINTING, OBSERVATION_INDEX), and getting all descriptions complete and correct and consistent is almost impossible if it's copy & pasted four times.

@jknodlseder @cboisson @TarekHC @lmohrmann - If you have time, please review this PR and leave a comment.

### DIFF
--- a/source/general/time.rst
+++ b/source/general/time.rst
@@ -5,46 +5,88 @@
 Time
 ====
 
-This page gives background information on times in gamma-ray astronomy.
-
-It's not a format specification, rather a summary of the status quo:
-
-* How times are stored in files.
-* How times are represented in science tool codes
-* How times are input by users and output to users from these codes.
-
-.. _time-introduction:
-
 Introduction
 ------------
 
-Times are used in many places in high-level analysis, e.g.
+This page describes how times should be stored.
+This is a solved problem, we follow the `FITS standard`_.
+However, the FITS standard is very complex (see also `FITS time paper`_),
+and allows for different ways to store times in FITS, some of which
+are hard to understand and implement.
 
-* Observations have start and end times and sometimes are split up into
-  "good time intervals" GTIs when hardware issues occur or clouds pass the field of view.
-* Gamma-ray events are observed at given times, and those times are needed
-  to convert the reconstructed AltAz position to RaDec, or to select events in
-  a given GTI.
-* Some gamma-ray sources are variable, e.g. AGNs can flare on timescales of seconds or minutes,
-  or pulsars emit a periodic signal on timescales of seconds or milli-seconds.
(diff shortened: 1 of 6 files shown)


lmohrmann:
@cdeil I don't have time to look at this in detail, but I'm certainly in favor of cleaning up the docs a bit. Your proposed changes look fine to me at first sight (but, as I said, I haven't looked at this in detail).

cdeil:
Anyone else wants to review this? If not I'll merge this in the next days.

source/general/time.rst
```diff
@@ -5,46 +5,88 @@
...
+* ``DATE-END`` type: string
+    * Observation end date (format: "yyyy-mm-dd")
+* ``TIME-END`` type: string
+    * Observation end time (format: "hh:mm:ss.sss...")
 
-Basically we follow `Time in Fermi data analysis`_, so this is the number one
-reference.
+Note that the FITS standard allows and it is quite common to instead put a
+``TIME-OBS`` key with value "yyyy-mm-ddThh:mm:ss.sss..." and to omit the ``DATE-OBS`` key
+(see `Dictionary of Commonly Used FITS Keywords`_). That is allowed as well.
```
TarekHC: I would rephrase this paragraph a bit (is not too formal at the moment). At least I would remove the "That is allowed as well".

cdeil:
@TarekHC - Thanks for the review. I've done some improvements in f86f7ab .

One suggestion I didn't follow was to add a subsection named "header keywords" to the section "Time format", because it didn't seem useful to have a more complex header structure for that page and to have a section with a single sub-section.

I did add a little more information on columns and header keys.

I'm merging this now.

If anyone finds an issue with this description, or has a suggestion to make it better, please leave a comment here, or better yet: open a pull request.

PR #106, opened 2018-04-23
Clarify normalisation of IRFs

lmohrmann:
This PR fixes a typo and explicitly states the normalisation requirements for the energy dispersion and PSF IRFs.

### DIFF
--- a/source/irfs/irf_components/index.rst
+++ b/source/irfs/irf_components/index.rst
@@ -45,15 +45,20 @@ Energy Dispersion
 -----------------
 
 The energy dispersion information is stored in a FITS file with one required
-extensions (HDU). The stored quantity is a PDF for the **energy migration**
+extensions (HDU). The stored quantity is :math:`\frac{dP}{d\mu}`, a PDF for the **energy migration**
 
 .. math::
 
      \mu = \frac{E_{\mathrm{reco}}}{E_{\mathrm{true}}}
 
-as a function of true energy and offset. It should be normalized to unity. The
-migration range covered in the file must be large enough to make this possible
-(Suggestion: :math:`1/3 < \mu < 3`)
+as a function of true energy and offset. It should be normalized to unity, i.e.,
+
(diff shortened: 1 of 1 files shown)


source/irfs/irf_components/index.rst
```diff
@@ -118,8 +123,7 @@ PSFs must be normalised to integrate to total probability 1, i.e.
 
 .. math::
 
-    \int_{0}^{\infty} 2 \pi r dP/dr(r) dr = 1, where dP/dr = 2 \pi r dP/d\Omega
-
+    \int_{0}^{\infty} dP/d\Omega(r) d\Omega = 1\,.
```
cdeil: @lmohrmann - Omega is the solid angle, I don't think the integral as written is correct.
either you put a generic `\int dP/dOmega dOmega` with the understanding that it's a 2D integral, or you have to write an integral over the 1D radial `r`, no?

lmohrmann: True, I removed the bounds.

source/irfs/irf_components/index.rst
```diff
@@ -104,7 +109,7 @@ Probability distributions
   This is the canonical form we use and the values we store in files.
 * Often, when comparing observered event distributions with a PSF model,
   the :math:`dP/dr^2` distributions in equal-width bins in :math:`r^2` is
-  used. The relation is :math:`d\Omega = \pi dr^2`, i.e. :math:`dP/dr^2=(1/\pi)(dP/d\Omega)`.
+  used. The relation is :math:`d\Omega = \pi dr^2`, i.e. :math:`dP/d\Omega=(1/\pi)(dP/dr^2)`.
```
cdeil: Just to point it out, since it's non-obvious: there was an error here in the relation between dP/dOmega and dP/dr^2, fixed by @lmohrmann in this PR. I have double checked that now all equations for the PSF are correct.

PR #108, opened 2018-06-20
Observatory location keys for headers

cdeil:
@jknodlseder @kosack all - How should we store observatory location in the FITS header?

What we mention currently is `GEOLON`, `GEOLAT` and `GEOALT` in the `POINTING` spec:
https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/blame/48988405270a433a89ec6de19c1f4bc0db4bc417/source/events/pointing.rst#L46

For the `EVENTS`, in HESS and for the CTA 1DC files we use `GEOLON`, `GEOLAT` and `ALTITUDE`, e.g. for HESS:
```
GEOLAT  = -23.2717777777778 / latitude of observatory (deg)
GEOLON  = 16.5002222222222 / longitude of observatory (deg)
ALTITUDE=            1835. / altitude of observatory (m)
```
Our spec says nothing what to put for `EVENTS`, but I think we should be consistent with `POINTING`.

The [FITS standard](https://fits.gsfc.nasa.gov/standard40/fits_standard40aa.pdf) only mentions `OBSGEO-X`, `OBSGEO-Y`,  `OBSGEO-Z`, not the keys we use.

Looking around, it seems different telescopes / missions use different keys.

Thoughts?

### DIFF
--- a/source/events/events.rst
+++ b/source/events/events.rst
@@ -98,6 +98,7 @@ Mandatory header keywords
 -------------------------
 
 The standard FITS reference time header keywords should be used (see :ref:`time-formats`).
+An observatory Earth location should be given as well (see :ref:`coords-location`).
 
 * ``HDUCLASS`` type: string
     * Signal conformance with HEASARC/OGIP conventions (option: 'OGIP'). See :ref:`hduclass`.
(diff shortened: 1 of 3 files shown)


lmohrmann:
:+1:

cdeil:
Merging now.
As always, this spec is work in progress; comments always welcome.

PR #13, opened 2018-06-22
Field of view coordinates and IRF axis names

cdeil:
The spec for field of view coordinates still needs to be written:
http://gamma-astro-data-formats.readthedocs.org/en/latest/info/coordinates.html#field-of-view

Also, I've started to write up recommended names for IRF coordinate axes:
http://gamma-astro-data-formats.readthedocs.org/en/latest/irfs/index.html#irf-axes
This info should probably be moved onto the coordinates page.

The hardest part is to describe / fix the FOV coordinates and names.
I'll make a proposal in the next few days and then we can discuss ...

### DIFF
--- a/source/events/events.rst
+++ b/source/events/events.rst
@@ -45,31 +45,24 @@ Optional columns
 
 * ``EVENT_TYPE`` tform: ``32X``
     * Event quality partition.
+* ``MULTIP`` tform: ``1I``
+    * Telescope multiplicity. Number of telescopes that have seen the event.
 * ``GLON`` tform: ``1E``, unit: deg
     * Reconstructed event Galactic longitude (see :ref:`coords-galactic`).
 * ``GLAT`` tform: ``1E``, unit: deg
     * Reconstructed event Galactic latitude (see :ref:`coords-galactic`).
 * ``ALT`` tform: ``1E``, unit: deg
-    * Reconstructed altitude coordinate of event
-      (horizon system, see :ref:`coords-altaz`)
+    * Reconstructed altitude (see :ref:`coords-altaz`)
 * ``AZ`` tform: ``1E``, unit: deg
-    * Reconstructed azimuth coordinate of event
-      (horizon system, see :ref:`coords-altaz`)
+    * Reconstructed azimuth (see :ref:`coords-altaz`)
 * ``DETX`` tform: ``1E``, unit: deg
-    * Reconstructed event X-coordinate in detector system
-      (nominal system, see :ref:`coords-fov`).
-      See also the `OGIP event list`_ standard.
+    * Reconstructed field of view X (see :ref:`coords-fov`).
 * ``DETY`` tform: ``1E``, unit: deg
-    * Reconstructed event Y-coordinate in detector system
-      (nominal system, see :ref:`coords-fov`).
-      See also the `OGIP event list`_ standard.
+    * Reconstructed field of view Y (see :ref:`coords-fov`).
 * ``THETA`` tform: ``1E``, unit: deg
-    * Reconstructed offset from the observation pointing position
+    * Reconstructed field of view offset angle (see :ref:`coords-fov`).
 * ``PHI`` tform: ``1E``, unit: deg
-    * Reconstructed position angle from the observation pointing position
-      (position angles are counted counterclockwise from celestial North)
-* ``MULTIP`` tform: ``1I``
-    * Telescope multiplicity. Number of telescopes that have seen the event
+    * Reconstructed field of view position angle (see :ref:`coords-fov`).
 * ``DIR_ERR`` tform: ``1E``, unit: deg
     * Direction error of reconstruction 
 * ``ENERGY_ERR`` tform: ``1E``, unit: TeV
--- a/source/general/coordinates.rst
+++ b/source/general/coordinates.rst
@@ -131,46 +131,75 @@ Astropy, i.e. this is without refraction corrections):
 Field of view
 -------------
 
-In Gamma-ray astronomy, sometimes field of view (FOV) coordinates are used.
-Specifically some :ref:`background models <bkg>` are in the FOV coordinate system
-and FOV coordinates can also be used for other IRFs.
+FOV coordinates are currently used in two places in this spec:
 
+1. Some :ref:`background models <bkg>` are in the FOV coordinate system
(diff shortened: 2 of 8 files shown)


cdeil:
@lmohrmann - This issue is the last substantial topic to finalise the v0.2 of the spec, that we need for the HESS release. I'd like to propose again to drop ``DETX`` and ``DETY`` from the HESS events table. IMO it's just a waste of space, and given that we don't have agreement on it's definition (we put ALTAZ aligned, ctools assumes RADEC aligned) it just creates confusion / problems. Both Gammapy and ctools work perfectly fine with EVENTS that don't have a DETX / DETY column.

@TarekHC - Please check what you fill in ``DETX`` / ``DETY`` for MAGIC and also consider removing it, or document the orientation clearly.

@mackaiver - I see that for FACT you're already not adding these two columns to your DL3 EVENTS files. Well done!

Please review the attached commit, and directly edit via a follow-up commit if something is incorrect or unclear.

For now, I just tried to clearly describe the different FOV coordinates in use, but leave the decision what to use open, stating that this will be made by CTA hopefullly soon (cc @kosack from CTA / ctapipe).

@jknodlseder - Maybe you want to comment on this one as well?

source/events/events.rst
```diff
@@ -45,31 +45,24 @@ Optional columns
...
 * ``DETX`` tform: ``1E``, unit: deg
-    * Reconstructed event X-coordinate in detector system
-      (nominal system, see :ref:`coords-fov`).
-      See also the `OGIP event list`_ standard.
+    * Reconstructed field of view X (see :ref:`coords-fov`).
 * ``DETY`` tform: ``1E``, unit: deg
-    * Reconstructed event Y-coordinate in detector system
-      (nominal system, see :ref:`coords-fov`).
-      See also the `OGIP event list`_ standard.
+    * Reconstructed field of view Y (see :ref:`coords-fov`).
```
lmohrmann: Shouldn't DETX and DETY be replaced by the more clearly named FOV_ALTAZ_LON, etc...?

source/irfs/full_enclosure/bkg/index.rst
```diff
@@ -58,7 +58,7 @@ Required columns:
 * ``ENERG_LO``, ``ENERG_HI`` -- ndim: 1, unit: TeV
     * Reconstructed energy axis
 * ``DETX_LO``, ``DETX_HI``, ``DETY_LO``, ``DETY_HI`` -- ndim: 1, unit: deg
-    * Field of view coordinates binning, see :ref:`coords-fov`
+    * Field of view coordinates binning (see :ref:`coords-fov`)
```
lmohrmann: These are of course also somewhat ambiguous, it's not clear along which system these are aligned. Add a note that this should be specified clearly?

lmohrmann:
@cdeil Once we have a decision here, I can remove `DETX` and `DETY` from the event columns in the HAP exporter.

cdeil:
> I'm OK with dropping DETX and DETY from the event column in general. Are you 100% sure it's not being used by ctools at the moment though?

They might be accessed and used when using a bkg3d model. I don't know. Can you or someone in Erlangen try? (I don't have ctools installed and don't want to spend time on it).

lmohrmann:
> They might be accessed and used when using a bkg3d model. I don't know. Can you or someone in Erlangen try?

We'll have a look

source/general/coordinates.rst
```diff
@@ -131,47 +131,76 @@ Astropy, i.e. this is without refraction corrections):
...
+different from the IACT camera coordinate detectors), it wasn't clear what to put,
+both ``(DETX, DETY) = (FOV_ALTAZ_LON, FOV_ALTAZ_LAT)``
+and ``(DETX, DETY) = (FOV_RADEC_LON, FOV_RADEC_LAT)``
+and very early on even TAN projections were used.
+
+Given this situation that there is no concensus yet, one suggestion is to avoid
+putting FOV coordinates in EVENTS, or if they are added, to clearly state how
+they are defined. This still leaves the problem with the background models,
+in case they are non-radially symmetric. We expect CTA to make a decision and to define
+FOV coordinate systems soon, to resolve this issue.
```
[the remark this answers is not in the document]
cdeil: I'm -1 on encouraging the use of new column names such as FOV_ALTAZ_LON, FOV_ALTAZ_LAT and starting to add support for those in exporters and Gammapy / ctools.
We weren't able to agree with DETX / DETY even while no-one was using it, so it's highly unlikely that we're going to agree on new column names.
I really suggest to leave this decision to the future and to CTA, we can't resolve this here while the relevant people aren't participating in the discussion and no-one has decision making power.

source/events/events.rst
cdeil: See my comment above.

source/irfs/full_enclosure/bkg/index.rst
cdeil: I have written long comments / notes in the general FOV section. Does it really help to partly re-hash those comments here?
If you think so, would you be willing to add a commit to this branch with any text changes you like?

cdeil:
Just to comment: on the Gammapy side, the plan is to only access RA / DEC / TIME and to compute FOV coordinates from that, and then to interpret background model DETX / DETY with ALTAZ alignment, i.e. how we fill it at the moment in HESS.
And then once CTA makes a decision (e.g. in 2019), change the supported format and coordinate definitions to that.

lmohrmann:
@cdeil It appears ctools does use `DETX` and `DETY`, albeit only to compute offset angles that are used to query the IRFs event-wise. Maybe someone involved in ctools development could verify this? As I said, I'd be OK with removing these columns if none of the tools uses them. I'd still vote for including both the Ra/Dec and Alt/Az aligned FOV coordinates in the HESS test release, but that's not at issue here, right?

source/general/coordinates.rst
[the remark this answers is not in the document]
lmohrmann: OK, fine with me!

source/irfs/full_enclosure/bkg/index.rst
lmohrmann: Agreed, this is probably not necessary. Ignore my comment.

cdeil:
@dtiziani confirmed that ctools runs fine without `DETX` / `DETY` present, so I'll remove it from the HESS release files EVENTS tables.

Merging this PR.

PR #25, opened 2018-07-21
Update about and how to contribute page

cdeil:
The about and how to contribute page should be updated:
http://gamma-astro-data-formats.readthedocs.org/en/latest/info/about.html

Add link to https://lists.nasa.gov/mailman/listinfo/open-gamma-ray-astro
and explain that high-level discussion should go there, detailed discussion in Github issues.

Also, there's people that want to read a PDF and send feedback on that and not use Github,
so we have to point out where to get the latest stable and master PDF version and where to send the comments.

### DIFF
--- a/source/about.rst
+++ b/source/about.rst
@@ -5,36 +5,52 @@
 About
 =====
 
-What is this?
--------------
+In gamma-ray astronomy, a variety of data formats and proprietary software have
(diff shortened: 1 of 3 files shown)


cdeil:
I have re-written the about / intro page.
Merging this now.

PR #130, opened 2018-09-17
Add RA_PNT and DEC_PNT as required header keys in EVENTS

cdeil:
This PR adds back RA_PNT and DEC_PNT as required header keys in EVENTS .

They were initially in, but then removed already in 2016 in #39 .

There was a lot of discussion on that point there, and I think in the end there was agreement that as long as POINTING isn't used much, those header keys should remain, and it was my oversight that they got lost.

@jknodlseder - OK?

@mackaiver - Please add those header keys for your FACT DL3 FITS writer. I think otherwise your data will not work with ctools, and will stop working with Gammapy soon when we change from accessing that info from the obs index files to accessing it from the EVENTS header.


cdeil:
I don't think this is controversial, it was just an oversight. Merging this now.

Of course, comments and suggestions still welcome here or in a new issue any time.

PR #139, opened 2020-03-18
Add FACT to possible values of TELESCOP

MaxNoe:


TarekHC:
Hi Max. I don't expect this to lead to any discussion... So we can merge. Although note the listed experiments were just examples, and not the only choices possible. So we probably don't want to list there all possible experiments in the future.

MaxNoe:
> Hi Max. I don't expect this to lead to any discussion... So we can merge. Although note the listed experiments were just examples, and not the only choices possible

This came up as a joke in the FACT telcon and I thouight I'd use it to demonstrate that this is an evolving sthandar :wink:

TarekHC:
Then you demonstrated your point! :)  It is (and needs to be) an evolving standard.

PR #140, opened 2020-03-19
Remove effective area as function of reco energy, fixes #137

### LINKED ISSUE #137: Effective area in estimated energy: How?
The website states: 

> In the case of using the effective area as a function of reconstructed energy, results may be less precise, but require less MC statistics and allows to perform faster likelihood fits by science tools.

How can you express effective area in estimated energy when effective area is the ratio of selected events of all simulated events and there is no estimated energy for the events that did not trigger / were not selected by a pre-selection / did not survive image cleaning?

MaxNoe:
This removes all mentions of effective area as a function of
reconstructed energy as this is not possible to calculate.

The assumption of unity energy migration can of course still be done,
but that changes nothing concerning the fact that effective area can
only ever be expressed in bins of true, simulated energy.

It also gives a bit more context on what effective area actually means.


lmohrmann:
This looks like a pretty reasonable request to me. 

Maybe just to be sure:
@adonath @registerrier - I think the effective area as a function of reconstructed energy (if present in AEFF FITS files) is never used in Gammapy. Can you confirm this?

registerrier:
Indeed, effective area is always defined as a function of true energy.

MaxNoe:
ctools is also pretty clear on this here: http://cta.irap.omp.eu/ctools/users/user_manual/irf_cta.html

In that notation, E' is reconstructed energy and E is true energy.

Pinging @jknodlseder

MaxNoe:
> The instrument response functions provide a mathematical description that links the measured quantities of an event to the physical quantities of the incident photon.

This!

adonath:
@MaxNoe No objections from my side, seems pretty clear and uncontroversial. I'm not sure what the motivation for this specification was in first place...maybe @TarekHC still knows?

Probably we should also adapt the example fits file and corresponding script? https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/blob/master/source/irfs/full_enclosure/aeff/create_example_file.py

Internally in Gammapy we always use effective area in true energy and handle the case of no energy dispersion matrix equivalent to the unity one, exactly as you suggested in your initial post.

MaxNoe:
I thought I checked the example file and it only contained aeff vs true energy. Hmm, let me check again.

MaxNoe:
You are right, I'll update the examples

GernotMaier:
I think this comes from the more classical codes which used instead of forward- or unfolding the simple correction method for energy reconstruction.  This method uses A(E_rec), which is accurate enough for most applications.

MaxNoe:
>  This method uses A(E_rec)

I don't understand how you can properly define an effective area in versus estimated energy.
How would that work?

GernotMaier:
This is a method commonly called 'method of correction factors' in statistics books.

- bins of true and data distribution needs to be identical
- assumption is that a good estimator for the true value in each bin is simply a correction factor times the measured value.

The correction factors are determined by MC in kind of exactly the same way we calculate effective areas: count the simulated events per bin (in true energy) and count the reconstructed events per bin (in reconstructed energy).  This way the spillover from bin-to-bin is taken into account.

Of course it is jargon to call this A(E_rec).

Issue #149, opened 2020-03-30
License


MaxNoe:
This project should have a license.

As it is not directly code, CC would probably most approrpiate.

As it is a standard, it should not be modified.

So I would propose CC by Attribution, No-Derivatives:
https://creativecommons.org/licenses/by-nd/4.0/

lmohrmann:
This sounds like a good proposal to me. Thoughts by anyone else?

tibaldo:
Do you need to verify that the format license does not conflict with the licenses of the any Science Tools using the format?

lmohrmann:
Another thought: this is supposed to be a prototype for the CTA data format, although it will of course be for CTAO to decide which format they will use. Nevertheless, we should make sure that a license does not prevent CTAO from building upon this work. I'm not an expert on licenses, so I don't know if this could be a problem or not.

lmohrmann:
Also, see the previous issue #114

MaxNoe:
Ok, sorry for not checking before opening this one. Seems they came to the same conclusion. BSD for code and CC BY for the license text. I'll open a PR.

PR #155, opened 2020-04-14
Add creative commons license, fixes #114

### LINKED ISSUE #114: License?
I think we should maybe add a license to this `gamma-astro-data-formats` to make it clear what kind of use is and isn't allowed.

This might be important especially if CTA wants to continue here or copy from here as the starting point for their internal data format specification effort. My understanding is that currently, it's legally not clear what use is allowed, CTA would have to contact the contributors here and ask permission before e.g. copying the spec as starting point for CTA-internal documentation.

Looking around a bit, e.g. at https://choosealicense.com/non-software/ or https://creativecommons.org/choose/ or e.g. https://github.com/elegant-scipy/elegant-scipy and https://github.com/jakevdp/PythonDataScienceHandbook as mixed code / text projects, one option could be that we put this:

- Code is BSD-3. (currently we just have a few Python scripts to generate example files using Numpy, Astropy and Gammapy, all BSD-3 projects)
- Text is [CC BY 4.0](https://creativecommons.org/licenses/by/4.0/).

These are very liberal licenses that are very permissive in re-use, even commercial.

@woodmd and @TarekHC  as biggest contributors here - OK?
@kosack  or anyone from CTAO - does this work for you?

MaxNoe:


lmohrmann:
I'd like to have some more people review this before merging.
@kosack @jknodlseder ?

Another thought that I already posted [here](https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/issues/149#issuecomment-606431374): we should probably make sure that CTAO can adopt this standard if they want to. As I'm not a license expert, I don't know if this could be a problem or not.

jknodlseder:
Here is what the _CTA Software Licensing Policy_ document says on document licenses:

_The CTA Observatory will always agree to a Creative Commons Attribution 4.0 International License [2] for documentation (provided there are no compatibility issues). Same applies for used icons, images, and other visual representations._

MaxNoe:
@jknodlseder Ok, great. Where could I find this document.

jknodlseder:
> @jknodlseder Ok, great. Where could I find this document.

Ask Igor Oya or Stefan Schlenstedt for a copy of the _CTA Software Licensing Policy_ document.

lmohrmann:
OK, I don't see any problems in adding this license then, and it seems that no one else does. Merging now.

PR #158, opened 2020-04-14
Change observatory according to with FITS standard, fixes #148

### LINKED ISSUE #148: Use observatory location definition as defined in the FITS standard
Section 9.3.2 reads:

> In order to provide a complete description, ’TOPOCENTER’ requires  the  observatory’s  coordinates  to  be  specified.  
> Thereare  three  options: 
> * (a) the  ITRS  Cartesian  coordinates  defined in  Sect.  8.4.1  (OBSGEO-X,OBSGEO-Y,OBSGEO-Z),  which  are **strongly  preferred**;
> * (b) a  geodetic  latitude/longitude/elevationtriplet (defined below);
> * (c) a reference to an orbit-ephemeris file.
> 
>  A set of geodetic coordinates is recognized by the followingkeywords. 
> * OBSGEO-B  [floating-point] The  value field  of  this  keyword shall contain the  latitude  of  the  observation  in  deg,  with North positive.
> * OBSGEO-L  [floating-point] The  value field  of  this  keyword shall contain the longitude of the observation in deg, with East positive.
> * OBSGEO-H  [floating-point] The  value field  of  this  keyword shall contain the altitude of the observation in meters.
> An orbital-ephemeris file can instead be specified. 
> OBSORBIT  [string] The value field of this keywordshallcon-tain the character-string URI, URL, or the name of an orbit-ephemeris file

We should probably replace GEOLON / GEOLAT / ALTITUDE

MaxNoe:


MaxNoe:
@lmohrmann done

lmohrmann:
@adonath @registerrier Does this change require some adaptions in Gammapy? I guess the tools should support both the old and new keywords for the time being.

PR #150, opened 2020-03-30
Add travis integration

MaxNoe:
* use poetry to define dependencies (sphinx and sphinx-rtd-theme at the moment)
* Build on travis
* deploy to gh-pages

PR #159, opened 2020-04-16
Add link to gh pages in README

MaxNoe:
After #150, we now have the current draft version deployed here:
https://open-gamma-ray-astro.github.io/gamma-astro-data-formats/

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,13 @@ For v0.2 we also archived the sources of the spec as well as a rendered PDF and 
 <https://doi.org/10.5281/zenodo.1409830> 
 To cite that version, you can use the bibtex entry [here](https://zenodo.org/record/1409831/export/hx#.W5EBLNgzY_U).
 
+## Draft version
+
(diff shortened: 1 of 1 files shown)


MaxNoe:
> The latest version?

The latest version? As in the current master branch? How?

lmohrmann:
Not automatically, but I can trigger a build. The changes you did in the last days should already be in.

MaxNoe:
Oh ok. How would that work, could this also be integrated with travis, so we don't deploy to gh pages but to readthedocs?

lmohrmann:
I've tried to set up a webhook to automatically trigger builds, but it seems that it doesn't work yet. I don't have time to investigate further now. Has anyone set up something like this for a repository already and can help?
For reference: I get [this message](https://docs.readthedocs.io/en/stable/webhooks.html#webhook-activation-failed-make-sure-you-have-the-necessary-permissions), but was unable to grant access to the ReadTheDocs OAuth App.

Issue #34, opened 2016-04-08
Event classes and tagging variables (as hadronness)


contrera:
**Proposal**: keep the format general enough so we can compute the flux of a source (or even a diffuse flux) either from the number of events surviving a cut in or from a fit to a “hadronness” distribution. I argue that this later method should be more performant at high contamination (low separation power) because in principle all events are kept and the whole information of the differences in hadronness pdfs is used. I also think this may be a road to work in cases where our separation power is small, like electron analysis.

**Use case**: we want to be able to compute the flux of a source using both methods

**Case 1: Using cuts**:
1.	Define a set of event classes corresponding to different cuts in hadronness 
2.	Compute  the effective area for each set of cuts Aeff(cut)
3.	Compute the background for each set of cuts from data N_bckg(cut)
4.	Count the number of events passing the cuts N_evts(cut)
5.	Compute the signal events N_sig(cut) = N_evts(cut)-N_bckg(cut)
6.	Flux(cut) = (N_sig(cut))/(Aeff(cut)*time)
Required: Class definitions, class for each event, Aeff(Cut), Backg(cut)

**Case 2: Fitting the distributions**:
1.	Keep the hadronness of each event
2.	Compute the pdf  for the signal hadronness_pdf(sig)
3.	Compute the pdf  for the background hadronness_pdf(bckg)
4.	Fit the hadronness distribution of the sample to
N_sig*hadronness_pdf(sig)+N_bckg*hadroness_pdf(bckg) 
5.	Flux  = N_sig/(Aeff*time)
 Required: hadroness for each evt, hadroness_pdf(evts), hadronness_pdf(bckg)
 
**Note:** A soft hadronness cut will be needed in any case to reduce the DL3 size

cdeil:
@contrera - Thank you for the proposal!

I think it would be nice to support more event info and those use cases in IACT DL3.
For me the way forward would be a pull request proposing changes and then discussing which extra info should be required or optional.

@jknodlseder @TarekHC, maybe @woodmd with the Fermi-LAT perspective - Thoughts?

TarekHC:
Hi all, 

[...]

PR #171, opened 2021-03-04
Added optional gammaness column in EVENTS hdu

cosimoNigro:
Shamelessly copying from what @maxnoe has just suggested in issue #34, a PR adding the optional `gammaness` column.
@maxnoe, from your suggestion I have replaced "this shower was induced by a gamma ray" with "the event was produced by a gamma ray", if we have to include gamma-ray satellites - though one can argue that what develops in a pair-conversion satellite is indeed a shower. But I thought it would be better to keep it more general.


maxnoe (approved):
Looks good to me

adonath (approved):
Thanks @cosimoNigro, no further comments from my side...

lmohrmann:
Thanks @cosimoNigro !

PR #172, opened 2021-04-07
Add rad max to possible HDU index values

maxnoe:
The rad max tables were missing from the HDU index table description.


TarekHC (approved):
Everything seems ok to me. I don't see any reason not to merge, as it is something that was already discussed and included into the specs.

Issue #174, opened 2021-04-23
PSF Table documentation


GernotMaier:
The documentation of the PSF tables is not entirely clear and could be clarified by better wording or using other expressions. There is jargon used, which makes it hard to understand. This issue is to discuss improvements.

This is the page: https://gamma-astro-data-formats.readthedocs.io/en/latest/irfs/full_enclosure/psf/psf_table/index.html#psf-table

1. *Field of view offset axis*: the FOV does not have an offset and there is no FOV-offset axis. I guess what is meant is the *angular distance to centre of the array pointing direction*? I suggest to change this accordingly. The current explanation requires to first understanding the FOV coordinates.
2. *Offset angle from source position*: this is simply the distance between simulated and reconstructed direction? This can be calculated without assuming an astronomical source (e.g., from diffuse simulations, or from many 'sources' on a grid in the FOV). I suggest the much simpler expression  'angular distance between true and reconstruction event direction'  

For the FOV coordinates (https://gamma-astro-data-formats.readthedocs.io/en/latest/general/coordinates.html#coords-fov):
- a little sketch would help a lot! 
- using equator and pole is mathematical correct, but might be confusing to readers who think about celestial plots or Earth's equatorial plane.
- note that at least for some experiments, these are not spherical coordinates but coordinates in a plane tangential to the array pointing direction (which for small FOVs and is approximately the same)

maxnoe:
Agree with mostly everything set here.

> note that at least for some experiments, these are not spherical coordinates but coordinates in a plane tangential to the array pointing direction (which for small FOVs and is approximately the same)

This is the only thing I disagree with. The standard here basically prescribes using proper spherical coordinates, which is the right thing to do. Experiments using other coordinate systems should convert to spherical when exporting IRFs in these formats.

TarekHC:
I also agree with @GernotMaier. Current specifications are full of jargons and we should definitely improve the wording.

The best way to proceed is for @GernotMaier to create a pull-request with the proposed changes, so we can directly comment on them (this one should be easy, you already proposed good alternatives in your issue). If there is general consensus that the changes improve the wording use, then we just accept the pull request.

GernotMaier:
> This is the only thing I disagree with. The standard here basically prescribes using proper spherical coordinates, which is the right thing to do. Experiments using other coordinate systems should convert to spherical when exporting IRFs in these formats.

[...]

PR #175, opened 2021-04-23
Improve description of PSF axes

GernotMaier:
Proposal to clarify the meaning of the axes in the PSF histograms (see also issue #174)

### DIFF
--- a/source/irfs/full_enclosure/psf/psf_table/index.rst
+++ b/source/irfs/full_enclosure/psf/psf_table/index.rst
@@ -18,9 +18,9 @@ Columns:
 * ``ENERG_LO``, ``ENERG_HI`` -- ndim: 1, unit: TeV
     * True energy axis
 * ``THETA_LO``, ``THETA_HI`` -- ndim: 1, unit: deg
-    * Field of view offset axis (see :ref:`coords-fov`)
+    * Angular distance to centre of the array pointing direction (see :ref:`coords-fov`)
 * ``RAD_LO``, ``RAD_HI`` -- ndim: 1, unit: deg
-    * Offset angle from source position
+    * Angular distance between simulated and reconstructed direction
 * ``RPSF`` -- ndim: 3, unit: sr^-1
     * Point spread function value :math:`dP/d\Omega`, see :ref:`psf-pdf`.
 


source/irfs/full_enclosure/psf/psf_table/index.rst
```diff
@@ -18,9 +18,9 @@ Columns:
 * ``ENERG_LO``, ``ENERG_HI`` -- ndim: 1, unit: TeV
     * True energy axis
 * ``THETA_LO``, ``THETA_HI`` -- ndim: 1, unit: deg
-    * Field of view offset axis (see :ref:`coords-fov`)
+    * Angular distance to centre of the array pointing direction (see :ref:`coords-fov`)
```
maxnoe, line 21: AE vs BE? Do we have a guideline for that? I would say AE

maxnoe: Please keep "axis", as this refers to the axis definition of the matrix

source/irfs/full_enclosure/psf/psf_table/index.rst
```diff
@@ -18,9 +18,9 @@ Columns:
...
 * ``RAD_LO``, ``RAD_HI`` -- ndim: 1, unit: deg
-    * Offset angle from source position
+    * Angular distance between simulated and reconstructed direction
```
maxnoe, line 23: Add axis here to be consistent.

source/irfs/full_enclosure/psf/psf_table/index.rst
TarekHC: I would not say "Array", as this may be used by non-IACT instruments...

GernotMaier: But then some one should change https://gamma-astro-data-formats.readthedocs.io/en/latest/general/coordinates.html#coords-fov , as here it is explicitly said that these types of coordinates are used for IACTs and that the centre of the FOV is the centre of the array pointing direction.

maxnoe: @TarekHC I would also say that the issue of rewording everything so it also works e.g. for HAWC is a larger one and should probably be done by someone from the Water Cherenkov community.

maxnoe: There are too many places where this is currently an issue.

maxnoe:
It's also not mentioned that these columns are lower and upper bin edges respectively. Maybe we should write it like this?


```

Columns:

* ``ENERG_LO`` -- ndim: 1, unit: TeV
    Lower bin edges of the true energy axis
* ``ENERG_HI`` -- ndim: 1, unit: TeV
    Upper bin edges of the true energy axis 
* ``THETA_LO`` -- ndim: 1, unit: deg
    Lower bin edges of the axis in the angular distance to center of the array pointing direction axis (see :ref:`coords-fov`)
*  ``THETA_HI`` -- ndim: 1, unit: deg
    Upper bin edges of the axis in the angular distance to center of the array pointing direction axis (see :ref:`coords-fov`)
* ``RAD_LO``  -- ndim: 1, unit: deg
     Lower bin edges of the axis in angular distance between simulated and reconstructed direction axis
* ``RAD_HI`` -- ndim: 1, unit: deg
    Lower bin edges of the axis in angular distance between simulated and reconstructed direction axis
* ``RPSF`` -- ndim: 3, unit: sr^-1
   Matrix of point spread function values in the binning defined in the above columns :math:`dP/d\Omega`, see :ref:`psf-pdf`.
```

PR #179, opened 2021-07-14
Add support for time keywords for sky map axes

adonath:
This PR adds a small addition to support additional keywords for time based map axes, based on the same time format standards used elsewhere in GADF.

### DIFF
--- a/source/skymaps/index.rst
+++ b/source/skymaps/index.rst
@@ -84,7 +84,11 @@ Header Keywords
       lower and upper edges of each bin.  If there is a single element
       the column should be interpreted as the bin center.  For
       ``EBOUNDS`` or ``ENERGIES`` HDUs this keyword is optional.
-  
+
+For time based axes the additional keywords described in `time-formats`_ are
+allowed.
+
+
 Columns
 ~~~~~~~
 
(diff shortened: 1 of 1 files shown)


TarekHC (approved):
I don't see any possible conflict on adding this, as it refers to something we have already discussed/approved.

lmohrmann (approved):

source/skymaps/index.rst
```diff
@@ -84,7 +84,11 @@ Header Keywords
       lower and upper edges of each bin.  If there is a single element
       the column should be interpreted as the bin center.  For
       ``EBOUNDS`` or ``ENERGIES`` HDUs this keyword is optional.
-  
+
+For time based axes the additional keywords described in `time-formats`_ are
+allowed.
```
maxnoe, line 89: Change to required? I think if there is a time axis, we need the time header keywords, right?

maxnoe (changes requested):
I think we actually need to require the time header keywords for time axis

source/skymaps/index.rst
adonath: Yes, I agree. If there is a time axis it should specify absolute times. The previous specification allowed for a time axis with relative time without reference time specification, which is questionable.

maxnoe (approved):

adonath:
Thanks for the review @lmohrmann, @TarekHC and @maxnoe. I don't have merge rights here so please go ahead if you think it's ready...

Issue #192, opened 2022-09-20
Make a final release before handing over to VODF?


maxnoe:
Should we make a final 0.3 release with the current changes to accomodate wide-field instruments?

I would be in favor, as there are already files floating around using these changes and some support is implemented in gammapy.

These are changes since 0.2:

https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/compare/v0.2...master

maxnoe:
Then main changes are:

* Following the FITS standard for observatory location (OBSGEO-{X,Y,Z} or OBSGEO-{B,L,H}) instead of OGIP (GEOLAT, GEOLON, ALTITUDE)
* Changes regarding wide field instruments (OBS_MODE, changes from required to optional columns)
* Adding RAD_MAX in the HDU index table


To minimize the changes, we could also think about reverting the first one (OBSGEO-X vs GEOXXX).

maxnoe:
So my first proposal would be:

* revert the observatory location change
* make the 0.3 release

lmohrmann:
Why revert the observatory location change? So as not to invalidate files that have been produced without taking into account this change?

In any case, I would be supportive of doing a release, it's been a long time since we had one.

maxnoe:
> So as not to invalidate files that have been produced without taking into account this change?

Just to keep the amount of work in the science tools for supporting 0.3 (and supporting 0.2 and 0.3 at the same time)  minimal.

It's a hard breaking change since it requires different keywords and different data representation to go with the recommended mode.

maxnoe:
Maybe @adonath , @registerrier can comment for the gammapy side, @jknodlseder or @tibaldo for ctools

adonath:
Thanks @maxnoe!  I completely agree it is time for another v0.3 release here. In Gammapy we already added support for some of the additional meta data (`OBS_MODE` and `RAD_MAX` table) in-officially but it would be great to have that tagged by a version number. In Gammapy we do not have a scheme to support multiple versions of GADF in the API (this means users cannot chose the version), however in practice we already support both v0.2 and v0.3. I think this is fine from our side.

I guess this will affect some of the data stored `gammapy-data`, which already includes the additional meta data but not an updated version tag. So I'll CC @LauraOlivera and @cosimoNigro here.

maxnoe:
@adonath Yes, these "in-between" files should be updated to have the released version they follow.


[...]

PR #193, opened 2022-09-27
Revert "Change observatory according to with FITS standard, fixes #148"

### LINKED ISSUE #148: Use observatory location definition as defined in the FITS standard

maxnoe:
Reverts open-gamma-ray-astro/gamma-astro-data-formats#158

To have minimal changes required in the science tools and exporter software, I propose to revert this change before releasing the final GADF 0.3.

This change however should be included in VODF.

### DIFF
--- a/source/general/coordinates.rst
+++ b/source/general/coordinates.rst
@@ -211,35 +211,26 @@ When working with Alt-Az coordinates or very high-precision times,
 an observatory Earth location is needed. However, note that high-level
 analysis for most use cases does not need this information.
 
-The `FITS standard`_ specifies two possibilities for defining observatory location:
+The FITS standard mentions ``OBSGEO-X``, ``OBSGEO-Y``, ``OBSGEO-Z``
+header keys, and we might want to consider using those in the future.
 
-* ITRS Cartesian coordinates (defined in section 8.4.1, strongly prefered according to section 9.3.2)
-* Geodetic latitude / longitude / elevation
+For now, as of 2018, however, the existing IACT FITS data uses the
+following header keys, so their use is encouraged:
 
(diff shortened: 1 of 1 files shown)


lmohrmann (approved):
Thanks, I think this makes sense after the discussion in #192 .

adonath (approved):
Fine from my side as well. I think the change you revert is not required for any functionality, but mostly "cosmetic" so from the Gammapy side we'd be happy to not have to support it for now.

PR #195, opened 2022-11-04
Add sentence about OBS_MODE for IACTs

maxnoe:
There was some confusion regarding OBS_MODE in LST in the current dev version, has the OGIP defined values only talk about satellites.

I added a note that most IACT data should use POINTING, indiciating constant equatorial values for the observation.


adonath (approved):
Thanks @maxnoe! I agree it makes sense to clarify the behavior for IACTs.

PR #197, opened 2022-11-08
Fix doi in badge

maxnoe:


adonath (approved):
Thanks @maxnoe, looks good!

